PR #6, opened 2018-03-30
fixed grabber.js to handle network issues and node restarts

rkononov:


rkononov:
@treeder could you please review so I could deploy this one (a lot of improvements)

PR #22, opened 2018-05-21
redesign block/homepage

chovy:
- add sidebar navigation
- dark color scheme
- change font colors and link colors
- add rounded corners to input fields
- use svg logo

add margin to page wrapper

- account for new navbar on left

add colors to tx and address pages


chovy:
screenshot 2018-05-21 15 31 13

.env.sh
```diff
@@ -0,0 +1,8 @@
+export PORT=8000
+export MONGO_URI=mongodb://localhost:27017/blockDB
+export RPC_URL=https://testnet-rpc.gochain.io:443
```
rkononov: I think you shouldn't post real values here

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+    image: gochain/explorer:latest
+    command: "node app.js"
+    links:
+        - "mongodb"
+    ports:
+        - "80:80"
+    depends_on:
+      - mongodb
+    environment:
+          - RPC_URL=https://testnet-rpc.gochain.io:443
```
rkononov: same here use env variables

.env.sh
rkononov: or just remove this file from commit

docker-compose.yml
rkononov: or just remove this file from commit

Issue #28, opened 2018-05-25
Format numbers with commas


jmank88:
#21 #23 Fixed decimals but lost comma formatting:
![screenshot from 2018-05-24 20-46-13](https://user-images.githubusercontent.com/1194128/40521950-dd82da60-5f93-11e8-8d2f-3e8c2e0dd7ba.png)
Maybe we should have commas in block number too:
![screenshot from 2018-05-24 20-48-16](https://user-images.githubusercontent.com/1194128/40521948-dbece1aa-5f93-11e8-8181-f11f95c2f9fb.png)

PR #29, opened 2018-05-25
Format numbers with commas #28

rkononov:
Format numbers with commas #28

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -308,6 +308,13 @@ BlocksApp.filter('timeDuration', function() {
     return getDifficulty(hashes);
   };
 }) 
+.filter('balance', function() {
+    return function(val) {        
+        var parts = val.toString().split(".");
+        parts[0] = parts[0].replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+        return parts.join(".");
+    };
+  }) 
 .filter('teraHashes', function() {
     return function(hashes) {
         var result = hashes / Math.pow(1000, 4);
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -308,8 +308,9 @@ BlocksApp.filter('timeDuration', function() {
     return getDifficulty(hashes);
   };
 }) 
-.filter('balance', function() {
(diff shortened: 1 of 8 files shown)


jmank88:
Looks good. What about block number?

rkononov:
@jmank88 not sure, for me block numbers are IDs rather than numbers

jmank88:
What is the difference? They are strictly numeric IDs, which will quickly be in the millions, and we already do this elsewhere:
screenshot from 2018-05-25 12-29-02
screenshot from 2018-05-25 12-30-18

rkononov:
@jmank88 updated

public/js/main.js
```diff
@@ -308,6 +308,14 @@ BlocksApp.filter('timeDuration', function() {
...
   };
 }) 
+.filter('bigNumber', function() {
+    return function(val) {        
+        if (val == null) return val;
+        var parts = val.toString().split(".");
+        parts[0] = parts[0].replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+        return parts.join(".");
+    };
+  })
```
jmank88: Is there a built-in directive we can use? Like this one: https://docs.angularjs.org/api/ng/filter/number

jmank88: Can we adjust that? "Number of decimal places to round the number to. If this is not provided then the fraction size is computed from the current locale's number formatting pattern. In the case of the default locale, it will be 3."

Or use [`currency`](https://docs.angularjs.org/api/ng/filter/currency)?

Otherwise we lose the locale customization of `,` v.s `.`.

rkononov: we can adjust (didn't work for me) but how many decimal places we should have after .?
Currency round up to 2 digits by default and add currency sign.

I think it's easier to add locale customization - , or . if it make sense (decimal and thousands separators)

jmank88: Yeah I guess we'd have to tell it to display max decimals, then we're back to truncating trailing zeros.

PR #39, opened 2018-05-27
remove underline on navbar links

chovy:


public/css/layout.css
```diff
@@ -77,6 +77,7 @@
     display: flex;
     justify-content: flex-start;
     align-items: center;
+    text-decortion: none;
```
treeder: spelling

chovy:
fixed.

Issue #34, opened 2018-05-26
Display small values in different units


jmank88:
![screenshot from 2018-05-26 11-17-47](https://user-images.githubusercontent.com/1194128/40578268-895715d4-60d6-11e8-8097-04930001df92.png)
Gas price, at least, is commonly displayed in `gwei`. Maybe we could dynamically display small values in other units as well.

rkononov:
![image](https://user-images.githubusercontent.com/187434/40623232-943b2b64-62c7-11e8-9d8b-67e56d1f0a13.png)

PR #40, opened 2018-05-28
Display small values in different units #34

rkononov:
#34 
image


treeder:
I don't like this, then I have to figure out what the hell kwei is. I think we should stick the least units possible. People know gwei for gas price so we can leave that, then the regular unit GO (and maybe wei like you have in parentheses since it's the smallest unit). I don't think we should add more units than that.

jmank88:
Oh, yeah. Gas provided is actually unit-less anyways. It's just 'units of gas'.

rkononov:
updated
image

Issue #42, opened 2018-05-30
Get rid of gwei on tx value


treeder:
Should just be GO. 

![screen shot 2018-05-29 at 7 10 56 pm](https://user-images.githubusercontent.com/75826/40695022-285ea22e-6374-11e8-9468-8b7f805150ba.png)

jmank88:
It looks like it was losing precision too. Need to be careful of that.

PR #43, opened 2018-05-30
updated convertion to gwei function, removed gwei on tx value

rkononov:
#42 Had to implement divByPowerOfTen function which doesn't round big numbers but in same time strip trailing zeroes.
Unfortunately, I'm not familiar with unit tests for angular, maybe @chovy could help here


rkononov:
@chovy ?

PR #45, opened 2018-05-31
GCR support

rkononov:
Added GCR support (push in both storages now)

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -3,7 +3,7 @@ set -exuo pipefail
 
 user="gochain"
 image="explorer"
-
+gcr_project="gochain-core"
 # ensure working dir is clean
 git status
 if [[ -z $(git status -s) ]]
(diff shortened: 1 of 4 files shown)


release.sh
```diff
@@ -30,4 +30,16 @@ git push origin $version
...
-docker push $user/$image:latest
\ No newline at end of file
+docker push $user/$image:latest
+
+#Push docker images to GCR
+export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)"
+echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+sudo apt-get update && sudo apt-get install google-cloud-sdk -y
+gcloud auth activate-service-account --key-file=${HOME}/gcloud-service-key.json
```
jmank88: Can we use the built-in support instead of installing this from the release script?
https://circleci.com/docs/2.0/google-container-engine/#authenticate-to-googles-container-registry

rkononov: Tried that before, 
 - image: google/cloud-sdk
is not compatible with https://github.com/gochain-io/explorer/blob/master/.circleci/config.yml#L4

jmank88: Yeah, you have to use `docker` instead. I switched over gochain a while back: https://github.com/gochain-io/gochain/blob/master/.circleci/config.yml

Everything is faster in general, and it would allow us to skip the 'install docker' step too.

rkononov:
@jmank88 updated

Issue #48, opened 2018-06-02
Display checksummed addresses


treeder:
https://kb.myetherwallet.com/addresses/what-does-checksummed-mean.html

There's probably a method in web3 that does this.

PR #50, opened 2018-06-04
add checksummed adddress on address page

chovy:
refs #48


public/js/controllers/AddressController.js
```diff
@@ -9,6 +9,7 @@ angular.module('BlocksApp').controller('AddressController', function($stateParam
 
     $rootScope.$state.current.data["pageSubTitle"] = $stateParams.hash;
     $scope.addrHash = $stateParams.hash;
+    $scope.checksummedAddrHash = web3.utils.toChecksumAddress($stateParams.hash);
```
rkononov: can we calculate this on backend and then just send to frontend like everything else?

public/js/services/Web3Service.js
```diff
@@ -0,0 +1,20 @@
+var Web3Service = angular.module('Web3Service', [])
+    .service('web3', function () {
+        // Connect to Web3
+        try {
+            if (typeof web3 !== 'undefined') {
```
treeder: We'll want this to point to our RPC endpoints, probably best to pass in via env var since we have 3 networks running.

chovy: I don't think we can pass ENV vars to the front end in this app.

public/js/controllers/AddressController.js
chovy: Done. I moved it to backend.

chovy:
@treeder I updated PR with an api call to get environment variables and app config.

rkononov:
@chovy why we need web3 on frontedn since you are receiving everything from backend (checksummed adress) - or am I wrong here?

rkononov:
Let's keep PR clean - since we don't need this code let's remove that

chovy:
@rkononov done.

views/index.ejs
```diff
@@ -144,6 +154,7 @@
         <script src="/plugins/angularjs/plugins/angular-ui-router.min.js" type="text/javascript"></script>
         <script src="/plugins/angularjs/plugins/ocLazyLoad.min.js" type="text/javascript"></script>
         <script src="/plugins/angularjs/plugins/ui-bootstrap-tpls.min.js" type="text/javascript"></script>
+        <script src="/plugins/web3.min.js" type="text/javascript"></script>
```
rkononov: do we need this one?

views/index.ejs
```diff
@@ -152,6 +163,7 @@
         <script src="/js/filters.js" type="text/javascript"></script>
 
         <script src="/js/directives.js" type="text/javascript"></script>
+        <script src="/js/services/ConfigService.js" type="text/javascript"></script>
```
rkononov: And this?

Issue #33, opened 2018-05-26
Add more tx fee information


treeder:
Etherescan has a few different values here that are useful:

![screen shot 2018-05-26 at 9 16 07 am](https://user-images.githubusercontent.com/75826/40578246-7aea34e2-60c5-11e8-8a31-b1d210e90909.png)

In particular, the actual cost.

chovy:
How do I get these values? This is the only response I get right now:

```
Formatted JSON Data
{  
   "blockHash":"0x7eae7f8f327e2315201809b1f31116e815f7b1f9fcd94dd4dd71af22482d03a9",
   "blockNumber":141794,
   "from":"0x47f46cbc59820efce84c1a724d9332fc95e31a2e",
   "gas":276019,
   "gasPrice":"3373712147",
   "hash":"0x8fe31812ddf4208c850f7f1dfa0abf088738aeb8bbb8931c50044aafe9891470",
   "input":"0x",
   "nonce":298589,
   "to":"0x6a39b9f07175166b2ed7b7745d47470248966b15",
   "transactionIndex":0,
   "value":"0.000000000000000017",
   "v":"0xf4f5",
   "r":"0x4e0ffd3c917fb32c583a80c63f4ea517d554801000de5ef2fde6449ca9a1a705",
   "s":"0x4c0fb44790cea8e3d416542c6ff602090e65a7ca289da19559d2268c6155b256",
   "timestamp":1527836222,
   "isTrace":false
}
```

treeder:
Actual cost should be:

```
gasPrice * gas = actual cost in WEI (convert it to GO for display)
```

Assuming gas == gas used. 

Not sure where gas limit is then?

chovy:
How do I convert WEI to GO?

treeder:
https://web3js.readthedocs.io/en/1.0/web3-utils.html#fromwei

`fromWei(x, 'ether')`

PR #57, opened 2018-06-07
add row for actual gas cost in go

chovy:
refs #33

### DIFF
--- a/routes/web3relay.js
+++ b/routes/web3relay.js
@@ -45,6 +45,7 @@ exports.data = function (req, res) {
       } else {
         var ttx = tx;
         ttx.value = etherUnits.toEther(new BigNumber(tx.value), "wei");
+        ttx.actualGasCost = web3.fromWei(tx.gasPrice * tx.gas , 'ether');
         //get timestamp from block
         var block = web3.eth.getBlock(tx.blockNumber, function (err, block) {
           if (!err && block)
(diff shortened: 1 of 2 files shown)


routes/web3relay.js
```diff
@@ -45,6 +45,7 @@ exports.data = function (req, res) {
       } else {
         var ttx = tx;
         ttx.value = etherUnits.toEther(new BigNumber(tx.value), "wei");
+        ttx.actualGasCost = web3.fromWei(tx.gasPrice * tx.gas , 'ether');
```
jmank88: Is it safe to mulitply these? Or do we need to carefully construct a `BigNumber` first?

chovy: I'm not sure, that's what @treeder told me to do.

jmank88: Typical/minimum values for `gas` and `gasPrice` are  `21000` and `2000000000000000000`, so the result of this multiplication will be `>=4.2e+22` which would overflow a `uint64`. I'm not sure what javascript does here, but overflowing or falling back to floating point are both no good. You should be able to convert both values to `BigNumber` to do the multiplication safely, then call `etherUnits.toEther` with the resulting `BigNumber` just like the previous line.

jmank88: How would the math work? Javascript doesn't have arbitrary precision arithmetic built-in, does it?

I think that maintaining full precision during the computation (with `BigNumber`), and displaying it formatted via the `bigNumber` function are two separate things that both need to be done.

chovy: ok

chovy:
test please

PR #68, opened 2018-06-19
optimized grabber, added checking parent

r-kononov:
tested on testnet, seems working, no issues so far, db refilled and fixed

### DIFF
--- a/tools/grabber.js
+++ b/tools/grabber.js
@@ -127,7 +127,8 @@ var checkParentBlock = function (web3, blockData, recursively) {
         var parentBlockNumber = blockData.number - 1;
         Block.findOne({ number: parentBlockNumber }, function (err, b) {
             if (err) {
-                console.log("Cannot find block in db:", err)
+                console.log("Cannot find block in db:", err);
+                grabBlock(web3, parentBlockNumber, true);
             } else {
                 if (b) {
                     if (b && blockData.parentHash != b.hash) {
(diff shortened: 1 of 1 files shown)


tools/grabber.js
```diff
@@ -118,21 +122,49 @@ var grabBlock = function (web3, blockHashOrNumber, listening) {
...
 }
 
+var checkParentBlock = function (web3, blockData, recursively) {
+    if (blockData.number) {
+        var parentBlockNumber = blockData.number - 1;
+        Block.findOne({ number: parentBlockNumber }, function (err, b) {
+            if (err) {
+                console.log("Cannot find block in db:", err)
+            } else {
+                if (b) {
```
jmank88: else parent does not exist in the database? Shouldn't we fetch in that case as well?

rkononov: parent block is always there but added additional grabBlock call - everything could happens

tools/grabber.js
```diff
@@ -118,21 +122,52 @@ var grabBlock = function (web3, blockHashOrNumber, listening) {
...
+                                    if (b) {
+                                        checkParentBlock(web3, blockData, true);
+                                    }
+                                }
+                            });
+                        }
 
+                    }
+                } else {
+                    grabBlock(web3, parentBlockNumber, true);
```
jmank88: Can we either grab this one by `parentHash`, or do this check and fetch up top and then still execute the parent hash check afterwards?

Issue #72, opened 2018-06-25
Use eth_totalSupply rpc


jmank88:
Start using this dynamic, up to date source for total supply: https://github.com/gochain-io/gochain/pull/220

PR #74, opened 2018-06-28
Use eth_totalSupply rpc #72

r-kononov:


routes/web3relay.js
```diff
@@ -31,6 +31,23 @@ else
 var newBlocks = web3.eth.filter("latest");
 var newTxs = web3.eth.filter("pending");
 
+exports.totalSupply = function (req, res) {
+  web3.currentProvider.sendAsync({
+    jsonrpc: "2.0",
+    method: "eth_totalSupply",
+    params: ['0x0'],
```
jmank88: Is this requesting the total supply of block #0 every time? No params should default to 'latest', like on the console here:
screenshot from 2018-06-28 12-02-33

routes/index.js
```diff
@@ -47,11 +47,6 @@ function getConfig(req, res){
...
 }
 
-function getTotals(req, res){  
-
-  res.send("1000000000");
-}
-
 function getCirculating(req, res){  
 
   res.send("500000000");
```
jmank88: Circulating should be made dynamic also: `totalSupply - 500000000`

jmank88: Or the actual current balance of the initial alloc account(s) instead of the fixed value `500000000`.

routes/web3relay.js
jmank88: Oh, and the result is in `wei`, not `GO`, just like `getBalance`.

routes/index.js
r-kononov: how to get the actual current balance of the initial alloc account? is this available via API?

jmank88: Perhaps it would be clearest to have a separate function `getLocked()` which fetches this account's balance directly for mainnet: https://explorer.gochain.io/addr/0xf75b6e2d2d69da07f2940e239e25229350f8103f and returns 0 otherwise.

r-kononov: only one account?(then I suppose we should set it as env variable)
Maybe better to make separate RPC method in gochain?

jmank88: The funds are not actually 'locked' in any meaningful way other than that we promise to hold them, and it only applies to mainnet, so I think it's fine to have it coded in to explorer and I hope it is as simple as that single initial allocation account, but there may be others to include as well.

PR #87, opened 2018-08-09
Angular6

r-kononov:


treeder:
Approved, but let's keep it on testnet for a bit and use it there until it's ready for prime time before updating mainnet.

server/main.go
```diff
@@ -0,0 +1,278 @@
...
+		limit, _ = strconv.Atoi(limitS)
+	}
+
+	if skip < 0 {
+		skip = 0
+	}
+
+	if limit > 500 || limit <= 0 {
+		limit = 100
+	}
```
jmank88: Wouldn't setting `> 500` to `500` make more sense?

But why cap it at all? This prevents fetching an atomic snapshot of the data. I think enforcing this cap on the *front end* makes sense, so that the browser isn't overloaded, but maybe not on the backend.

server/main.go
```diff
@@ -0,0 +1,278 @@
...
+	err := app.Run(os.Args)
+	if err != nil {
+		log.Fatal().Err(err).Msg("Run")
+	}
+
+}
+func getTotalSupply(w http.ResponseWriter, r *http.Request) {
+	totalSupply, _ := backendInstance.TotalSupply()
+	totalSupply = new(big.Int).Div(totalSupply, big.NewInt(1000000000000000000)) //return in GO instead of wei
```
jmank88: These are rounding off to whole $GO. Should use big.Float or even big.Rat instead for better or complete precision.

server/models/address.go
```diff
@@ -0,0 +1,22 @@
+package models
+
+import "time"
+
+type Address struct {
+	Address                      string    `json:"address" bson:"address"`
+	Owner                        string    `json:"owner" bson:"owner"`
+	Balance                      string    `json:"balance" bson:"balance"`
```
jmank88: These values look like large integers now. Were the units changed from `GO` to `wei`? We should remain compatible with the old api, or use a new service path if we need to change types.

jmank88: I think dealing with the integers is cleaner though, and that is good to have in an additional new field.

server/models/address.go
```diff
@@ -0,0 +1,22 @@
...
+	BalanceInt                   int64     `json:"balance_int" bson:"balance_int"`
+	LastUpdatedAt                time.Time `json:"last_updated_at" bson:"last_updated_at"`
+	TokenName                    string    `json:"token_name" bson:"token_name"`
```
jmank88: Adding the `omitempty` json option to many of these fields would clean up and reduce the size of the response. Alternatively, they could be grouped into an object on a single field that would be null or omitted.

server/main.go
jmank88: Maybe something like this?
```go
var weiConvert = big.NewRat(1, 1000000000000000000)

func getTotalSupply(w http.ResponseWriter, r *http.Request) {
	totalSupply, _ := backendInstance.TotalSupply()
	total := new(big.Rat).Mul(new(big.Rat).SetInt(totalSupply), weiConvert) // return in GO instead of wei
	w.Write([]byte(total.FloatString(18)))
}

func getCirculating(w http.ResponseWriter, r *http.Request) {
	circulatingSupply, _ := backendInstance.CirculatingSupply()
	circulating := new(big.Rat).Mul(new(big.Rat).SetInt(circulatingSupply), weiConvert) // return in GO instead of wei
	w.Write([]byte(circulating.FloatString(18)))
}
```

server/models/block.go
```diff
@@ -0,0 +1,38 @@
+package models
```
jmank88: We should make these `models` types easily available for import into client projects. What do you think about moving them to the top level of the repo? That way they can import `github.com/gochain-io/explorer` and use `explorer.Block`, rather than `github.com/gochain-io/explorer/server/models` with `models.Block` (or an import alias).

server/models/address.go
```diff
@@ -0,0 +1,22 @@
+package models
+
+import "time"
+
+type Address struct {
+	Address                      string    `json:"address" bson:"address"`
+	Owner                        string    `json:"owner,omitempty" bson:"owner"`
+	Balance                      int64     `json:"balance" bson:"balance"` //backward compability
```
jmank88: This field comes through as a string fractional in the current version: `"balance":"18859853.03232"`

server/main.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func getAddress(w http.ResponseWriter, r *http.Request) {
+	addressHash := chi.URLParam(r, "address")
+	log.Info().Str("address", addressHash).Msg("looking up address")
+	address := backendInstance.GetAddressByHash(addressHash)
+	balance, err := backendInstance.BalanceAt(addressHash, "pending")
+	if err == nil {
+		address.BalanceWei = balance.String()                                                                         //to make sure that we are showing most recent balance even if db is outdated
+		address.Balance, _ = new(big.Float).Quo(new(big.Float).SetInt(balance), new(big.Float).SetInt(wei)).Float64() //converting to GO from wei
```
jmank88: Can we use `big.Rat` and construct the string ourselves so we don't loose any precision?
`big.Float` looses precision at ~20 digits (https://play.golang.org/p/7U4ZIY6DoWr), and float64 is even worse.

treeder: I was just dealing with this last week, but you can fix it by using `SetPrec(...)`, eg:

https://play.golang.org/p/q98GFdFA9DO

server/main.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func getAddress(w http.ResponseWriter, r *http.Request) {
+	addressHash := chi.URLParam(r, "address")
+	log.Info().Str("address", addressHash).Msg("looking up address")
+	address := backendInstance.GetAddressByHash(addressHash)
+	balance, err := backendInstance.BalanceAt(addressHash, "pending")
+	if err == nil {
+		address.BalanceWei = balance.String()                                                                                      //to make sure that we are showing most recent balance even if db is outdated
+		address.Balance, _ = new(big.Float).SetPrec(100).Quo(new(big.Float).SetInt(balance), new(big.Float).SetInt(wei)).Float64() //converting to GO from wei
```
jmank88: This is not sufficient - the precision is lost when converting to float64. Can we just use `big.Rat` like above to remain *exact*, and build the string ourselves with `FloatString(18)`?

PR #160, opened 2018-11-09
merge most recent changes to master

r-kononov:


r-kononov:
@bergauz could you please add something like "these counters updated every 5 minutes"

PR #164, opened 2018-11-23
Contract verification, closes #128

### LINKED ISSUE #128: Validating contracts
This sort of exists in the old explorer: https://explorer.gochain.io/addr/0x3b23f2485253c98303b8f9154e769345d83051fc#tab_addr_3

But apparently doesn't work. Can we add something like this to the new explorer?

bergauz:


server/backend/backend.go
```diff
@@ -90,6 +96,57 @@ func (self *Backend) GetBlockByHash(hash string) *models.Block {
...
+	}
+	// compiler gives map with keys starting with <stdin>:
+	key := "<stdin>:" + contractData.ContractName
+	if _, ok := compileData[key]; !ok {
+		err := errors.New("invalid contract name")
+		return nil, err
+	}
+	byteCodeFromSource := compileData[key].Code
+	// removing 0x
+	reg := regexp.MustCompile(`^0x`)
```
r-kononov: no need to use regexp here - just check first 2 symbols

server/backend/backend.go
```diff
@@ -90,6 +96,57 @@ func (self *Backend) GetBlockByHash(hash string) *models.Block {
...
+	finalCode := reg.ReplaceAllString(byteCodeFromSource, "")
```
r-kononov: same here, just replace 2 first symbols if needed

server/backend/db_backend.go
```diff
@@ -340,6 +340,19 @@ func (self *MongoBackend) importInternalTransaction(contractAddress string, tran
 	return internalTransaction
 }
 
+func (self *MongoBackend) importContract(contractAddress string, byteCode string) *models.Contract {
+	contract := &models.Contract{
+		Address:  contractAddress,
+		Bytecode: byteCode,
+	}
```
r-kononov: add created_at field

server/backend/db_backend.go
```diff
@@ -340,6 +340,19 @@ func (self *MongoBackend) importInternalTransaction(contractAddress string, tran
...
+	_, err := self.mongo.C("Contracts").Upsert(bson.M{"address": contract.Address, "code": contract.Bytecode}, contract)
```
r-kononov: upsert by address is fine (bson.M{"address": contract.Address}

PR #187, opened 2018-12-10
Redesign

bergauz:


server/models/block.go
```diff
@@ -27,6 +27,7 @@ type LightBlock struct {
 	CreatedAt time.Time `json:"created_at" bson:"created_at"`
 	Miner     string    `json:"miner" bson:"miner"`
 	TxCount   int       `json:"tx_count" bson:"tx_count"`
+	ExtraData string    `json:"extra_data" bson:"extra_data"`
```
jmank88: Only the first 32 bytes are intended for display. The rest will be vote information (not even UTF-8?). You could trim it on the server since `LightBlock` doesn't need vote info, or you could trim on the front end since the block details page could show vote info and share the logic to fetch the first 32 'vanity' bytes.

r-kononov: Oh thanks for pointing that, I'll update (I've checked but haven't noticed any issues - additional bytes or something)

jmank88: Here is an example from https://testnet-explorer.gochain.io/block/0x5b041d9147a8c7654f598379afe846af3ac88a649fd4d90d7a9ebf2d98540366
screenshot from 2018-12-10 07-50-09

Issue #184, opened 2018-12-08
Add filters into /api/


alexandrebelrouchi:
Hi guys,

Would be really nice to add filters such as Input data: 

like /api/address/transactions?input_data=XXXXXXX
like /api/address/transactions?created_at=XXXXXXX

Very useful for plenty of people I guess and would lighten up big time bandwidth cost as currently we need to load the whole DB to do it

Also, start fetching list from DESC/ASC

bergauz:
What about smth like this: 
POST:
{
  limit?: number,
  offset?: number,
  sort?: ASC | DESC,
  where?: {
    created_at: ''",
    other_fields...
  }
}
or GET: 
{
  limit?: number,
  offset?: number,
  sort?: ASC | DESC,
  created_at: '',
  other_fields...
}

r-kononov:
where?: {
created_at: ''",
other_fields...
}

bad Idea, we don't have indexes for every field and it doesn't make sense

jmank88:
@alexandrebelrouchi Can you be more specific about the queries you had in mind? 
Would `input_data` be a full match? Or a substring match? Or a regex?
Would `created_at` be exact? Or would before/after variants be necessary as well?

alexandrebelrouchi:
Input_data would be a regex if possible
Created at would be nice with a between X & X

alexandrebelrouchi:
> where?: {
> created_at: ''",
> other_fields...
> }
> 
> bad Idea, we don't have indexes for every field and it doesn't make sense

Yes but wouldn't that be better instead of fetching my 5000 transactions all the time?

r-kononov:
> > where?: {
> > created_at: ''",
> > other_fields...
> > }
> > bad Idea, we don't have indexes for every field and it doesn't make sense
> 
> Yes but wouldn't that be better instead of fetching my 5000 transactions all the time?
> Fetching 5k transactions every x seconds * 500 users for example would cost so much bandwidth ;)
> 
> I don't see how it doesn't make sense... way lighter on the backend

hmm we have 795,608,660 transactions atm and 13 fields - how much memory would that costs to have an index for every field and also it won't help for some of the regex queries (only for prefix ones), so I think even 1 user with a regex query could force full scan of the transaction collection which could lead to slowdown.

I'll add indexes and API for created_at_gt and created_at_lt and will test regexp/substring performance

r-kononov:
[...]

PR #199, opened 2018-12-26
Add filters into /api/ #184

r-kononov:


grabber/Gopkg.lock
```diff
@@ -20,7 +20,7 @@
   version = "v1.7.0"
 
 [[projects]]
-  branch = "optimize-holders"
+  branch = "extend-api"
   name = "github.com/gochain-io/explorer"
   packages = [
     "server/backend",
```
jmank88: It might be cleaner and more idiomatic to run `dep` from the top level in one place for everything - then these inter-repo reference go away.

PR #202, opened 2019-01-04
Wallet

bergauz:
Moved wallet to explorer


src/app/modules/wallet/wallet.service.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  get w3() {
+    return this._web3;
+  }
+
+  /**
+   * set mainnet rpc for mainnet explorer either testnet
+   */
+  static getHost(): string {
+    return /^explorer\.gochain\.io/.test(location.hostname) ? 'https://rpc.gochain.io' : 'https://testnet-rpc.gochain.io';
```
r-kononov: I think we should get this from the current domain (testnet-explorer/explorer)

src/app/services/api.service.ts
```diff
@@ -16,7 +16,7 @@ export class ApiService {
   apiURL: string;
 
   constructor(private http: HttpClient, private toastrService: ToastrService) {
-    this.apiURL = this.getApiURL();
+    this.apiURL = 'https://testnet-explorer.gochain.io/api' /*this.getApiURL()*/;
```
r-kononov: hardcoded?

inkberk: #202

PR #235, opened 2019-03-05
use internation date/time, closes #234

### LINKED ISSUE #234: Use International Date/Time Standard
I keep getting thrown off by ambiguous dates, like:
screenshot from 2019-03-02 09-39-27

We should use an international date/time standard like, [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) or [RFC 3339](https://tools.ietf.org/html/rfc3339).

bergauz:


jmank88:
Could probably swap the `T` for a space for readability if that is simple

jmank88:
Also, for blocks we only have precision to the second

jmank88:
As for whether we want to display in UTC or local time, I'm not sure, but it should be explicit whichever it is.

treeder:
UTC good enough to start.

jmank88:
> UTC good enough to start.

What do you mean to start? We only need to decide one way or the other. It shouldn't be any more work either way. The current behavior is local time. This PR changes to UTC, but perhaps only incidentally. If that is not what we want, then we shouldn't change it in the first place, instead of swapping back and forth.

jmank88:
A less disruptive 'to start' change would be to just swap `dd/MM/yyyy` for `yyyy/MM/dd`.

src/app/scenes/address/address.component.html
```diff
@@ -176,10 +176,10 @@ <h1>Explore your block</h1>
                       <td class="uk-text-truncate"><a class="ws-p" routerLink="/tx/{{tx.transaction_hash}}">{{tx.transaction_hash}}</a></td>
                       <td>
                         <ng-container *ngIf="tx.created_at != '0001-01-01T00:00:00Z';else updated_at">
-                          {{tx.created_at| date: 'dd/MM/yyyy HH:mm:ss'}} ({{tx.created_at | timeAgo }})
+                          {{tx.created_at}} ({{tx.created_at | timeAgo }})
```
jmank88: There are a lot of `... != '0001-01-01T00:00:00Z'` checks in the surrounding code. Unless that is some kind of angular trick, that looks like a very brittle comparison which depends on the output format. Is this the standard way of doing that? Or is there another way to check for null or 0 or something?

PR #237, opened 2019-03-14
Cleanup

jmank88:
This PR does a bunch of cleanup. The commits are best viewed separately.

- remove `OLD` dir
- reorganize makefile
- document cli commands
- remove redundant defaults in cli help
- update readme with help text in place of old docs
- make `backend_tests.go` compile and behind `integration` build tag
- run `go test -tags=integration` from integration test script
- make integration test script clean up after itself
- cache deps in docker file
- reformat readme json

The docs still need more work.


jmank88:
The backend tests fail on my local machine, and hang indefinitely on CI.

run_tests.sh
```diff
@@ -1,9 +1,14 @@
 set -e
 docker run --name test_mongo -d -p 8545:8545 -p 8080:8080 -p 27017:27017 mongo
+trap "docker stop test_mongo && docker rm test_mongo" EXIT SIGINT SIGTERM
+# TODO restore go test -tags=integration ./...
+go test ./...
```
jmank88: I will open an issue to fix the go integration tests. None of the go tests were running before, so this is still an incremental improvement for now.

jmank88: I moved this step in to the docker build (behind a `build-arg` switch) so we are building/caching in the same place.

jmank88: The integration tests should be able to call out to the existing mongo container from there (we already had it running anyways), but they are still off for now.

PR #244, opened 2019-03-25
replaced solc with docker image

r-gochain:
Replaced solc binary with docker image call, should be much easier to use web3 to compile and deploy or set the specific solc version in the sol header (not sure that selecting version from the list is a good idea - there are tons of versions - https://hub.docker.com/r/ethereum/solc/tags)

### DIFF
--- a/server/backend/solc.go
+++ b/server/backend/solc.go
@@ -1,7 +1,6 @@
 package backend
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
@@ -10,7 +9,6 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
-	"strconv"
 	"strings"
 )
 
@@ -44,8 +42,7 @@ type ContractInfo struct {
 
 // Solidity contains information about the solidity compiler.
 type Solidity struct {
-	Path, Version       string
-	Major, Minor, Patch int
+	Path, Version string
 }
 
 // --combined-output format
@@ -64,49 +61,19 @@ func (s *Solidity) makeArgs() ([]string, error) {
 		return nil, err
 	}
 	return []string{
-		"run", "-i", "-v", dir + ":/workdir", "-w", "/workdir", "ethereum/solc:" + strconv.Itoa(s.Major) + "." + strconv.Itoa(s.Minor) + "." + strconv.Itoa(s.Patch),
+		"run", "-i", "-v", dir + ":/workdir", "-w", "/workdir", "ethereum/solc:" + s.Version,
 		"--combined-json",
 		"bin,bin-runtime,srcmap,srcmap-runtime,abi,userdoc,devdoc,metadata",
 		"--optimize", // code optimizer switched on
 	}, nil
 }
 
-// SolidityVersion scan the source code and parse the version from pragma attribute
-func SolidityVersion(source string) (*Solidity, error) {
-	var err error
-	scanner := bufio.NewScanner(strings.NewReader(source))
-	for scanner.Scan() {
-		line := scanner.Text()
-		if strings.HasPrefix(strings.TrimSpace(line), "pragma solidity") {
-			matches := versionRegexp.FindStringSubmatch(line)
-			if len(matches) != 4 {
-				return nil, fmt.Errorf("can't parse a solc version %q", source)
-			}
-			s := &Solidity{Path: "docker", Version: matches[0]}
-			if s.Major, err = strconv.Atoi(matches[1]); err != nil {
-				return nil, err
-			}
-			if s.Minor, err = strconv.Atoi(matches[2]); err != nil {
-				return nil, err
-			}
-			if s.Patch, err = strconv.Atoi(matches[3]); err != nil {
-				return nil, err
-			}
-			return s, nil
-		}
-	}
-	return nil, fmt.Errorf("can't find a pragma solidity keyword")
-}
-
 // CompileSolidityString builds and returns all the contracts contained within a source string.
-func CompileSolidityString(ctx context.Context, source string) (map[string]*Contract, error) {
+func CompileSolidityString(ctx context.Context, compilerVersion, source string) (map[string]*Contract, error) {
 	if len(source) == 0 {
 		return nil, errors.New("solc: empty source string")
 	}
-	s, err := SolidityVersion(source)
-	if err != nil {
-		return nil, err
-	}
+	s := &Solidity{Path: "docker", Version: compilerVersion}
 	args, err := s.makeArgs()
 	if err != nil {
 		return nil, err
--- a/server/backend/solc.go
+++ b/server/backend/solc.go
@@ -43,8 +43,8 @@ type ContractInfo struct {
 
 // Solidity contains information about the solidity compiler.
 type Solidity struct {
-	Path, Version, FullVersion string
-	Major, Minor, Patch        int
+	Path, Version       string
+	Major, Minor, Patch int
 }
 
 // --combined-output format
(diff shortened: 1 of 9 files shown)


server/backend/solc.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	return []string{
+		"run", "-i", "-v", dir + ":/workdir", "-w", "/workdir", "ethereum/solc:" + strconv.Itoa(s.Major) + "." + strconv.Itoa(s.Minor) + "." + strconv.Itoa(s.Patch),
+		"--combined-json",
+		"bin,bin-runtime,srcmap,srcmap-runtime,abi,userdoc,devdoc,metadata",
+		"--optimize", // code optimizer switched on
+	}, nil
+}
+
+// SolidityVersion runs solc and parses its version output.
```
jmank88: Comment needs update?

server/backend/solc.go
```diff
@@ -0,0 +1,167 @@
...
+func SolidityVersion(source string) (*Solidity, error) {
+	var err error
+	matches := versionRegexp.FindStringSubmatch(source)
```
jmank88: This seems a little brittle. Can we find `pragma solidity` first to narrow the match to a single line?
I'm also concerned about how complex these matching rules can be. Since the user will need to be able to choose a specific version anyways, what about only allowing the simplest, single-version form `pragma solidity x.y.z;`? (and that is how they choose, rather than selecting from a drop down or whatever else)

server/backend/solc.go
```diff
@@ -0,0 +1,167 @@
...
+	s, err := SolidityVersion(source)
+	if err != nil {
+		return nil, err
+	}
+	args, err := s.makeArgs()
+	if err != nil {
+		return nil, err
+	}
+	args = append(args, "--")
+	cmd := exec.CommandContext(context.Background(), s.Path, append(args, "-")...)
```
jmank88: Might want to use a timeout at least (defense again spamming large/complex compilations), and maybe even wire the context from the user request.

server/backend/solc.go
```diff
@@ -0,0 +1,167 @@
...
+	SrcMapRuntime   string      `json:"srcMapRuntime"`
+	AbiDefinition   interface{} `json:"abiDefinition"`
+	UserDoc         interface{} `json:"userDoc"`
+	DeveloperDoc    interface{} `json:"developerDoc"`
+	Metadata        string      `json:"metadata"`
+}
+
+// Solidity contains information about the solidity compiler.
+type Solidity struct {
+	Path, Version, FullVersion string
```
jmank88: `FullVersion` unused?

server/backend/solc.go
jmank88: Actually, we don't even need to parse the version at all, do we? Can't we just take whatever is between the pragma and the next semi-colon, trim it, and leave it whole? `pragma solitidy <version>;`

server/backend/solc.go
```diff
@@ -0,0 +1,172 @@
...
+		"--optimize", // code optimizer switched on
+	}, nil
+}
+
+// SolidityVersion scan the source code and parse the version from pragma attribute
+func SolidityVersion(source string) (*Solidity, error) {
+	var err error
+	for _, line := range strings.Split(source, "\n") {
+		if strings.HasPrefix(strings.TrimSpace(line), "pragma solidity") {
+			matches := versionRegexp.FindStringSubmatch(line)
```
jmank88: I think this is still too strict (e.g. won't `pragma solidity 5.0;` fail to parse?), and potentially incorrect - it still ignores the potential complexity of multiple matching rules (e.g. they might be telling us explicitly which version(s) _not_ to use).
Once we have `pragma solidity`, we could just take whatever is between that and the next `;`, and use it in full (`Major`, `Minor`, and `Patch` are only used to recreate the original anyways, right?). This way we would never misinterpret them, and would require a strict single version selection by the user, which is exactly the docker image tag to use.

server/backend/solc.go
```diff
@@ -0,0 +1,172 @@
...
+		"--combined-json",
+		"bin,bin-runtime,srcmap,srcmap-runtime,abi,userdoc,devdoc,metadata",
+		"--optimize", // code optimizer switched on
+	}, nil
+}
+
+// SolidityVersion scan the source code and parse the version from pragma attribute
+func SolidityVersion(source string) (*Solidity, error) {
+	var err error
+	for _, line := range strings.Split(source, "\n") {
```
jmank88: It appears that newlines are just treated as whitespace and are not otherwise meaningful in solidity, so this should probably split on `;`. Also, something iterative, like a `bufio.Scanner/Reader`, would be mostly O(1) instead of O(n) since the convention is to put `pragma solidity` at the top.

server/backend/solc.go
r-gochain: @jmank88 
> take whatever is between that and the next ;
not a good idea because this could lead to command injection -ie  `docker run -it ethereum/solc:fake_version || some bad command here to run || the rest of the parameters`

So I'd prefer to leave as it is, but I could scan for additional symbols (<>^~) and require to set exact version instead.

jmank88: Validation still certainly make sense.

So if someone deploys with e.g. `nightly-0.5.0-f6f0cecc2f64f4fcda3da467bbe306b5859729aa`, we are just not going to support that?

jmank88: `pragma solidity` is actually not even required.

What is the problem with using a separate input to select the version?

jmank88: IMHO we should reconsider a separate input. We can do a best-effort auto-select based on this method, and let them choose whatever they need if it doesn't work. We can easily improve on the remix selector, by allowing for simple searching/auto-complete.

r-gochain: > Validation still certainly make sense.
> 
> So if someone deploys with e.g. `nightly-0.5.0-f6f0cecc2f64f4fcda3da467bbe306b5859729aa`, we are just not going to support that?

hmm I'd prefer not to support such cases, because it seems like there are differences between solc versions (https://github.com/ethereum/solidity/releases) and solc docker image versions (https://hub.docker.com/r/ethereum/solc/tags) so better to have have strict match

r-gochain: > `pragma solidity` is actually not even required.
> 
> What is the problem with using a separate input to select the version?

that's possible as well but maybe be separate ticket?(also including parsing the sol code on a client side and preselecting the proper version from the list)

jmank88: I don't think that is a good reason to abandon a subset of use cases. We can do better.

Can we at least match the versions available via Remix?
remix-solidity

r-gochain: I believe they compile solc by themselves, we use docker image, so yes we could show the list of docker image tags but it wouldn't be pretty useful (ie there is no 0.6.0 version in docker image)

r-gochain: I'd prefer user to use web3 to compile and deploy and then use explorer to verify (that why I vote for using the same approach in web3 and explorer)

jmank88: That is an unfortunate limitation, but why wouldn't it be useful? There are plenty of other intermediate versions that a user may need to produce the exact bytecode that they deployed. Why wouldn't we offer them that opportunity?

jmank88: We can't force them to use web3. We have to accommodate verifying deployed bytecode that was compiled by unknown manual processes or tools.

r-gochain: In this case it makes sense to implement the same feature in web3 to keep the same UX

r-gochain: @treeder do you want to allow user to select any solc version in both explorer and web3? or it should be both - trying to autoselect the solc version or user could force it?

jmank88: I'm all for aligning `web3` and explorer, but isn't the current approach broken - full stop?

A simple case: `pragma solidity <0.5.1;` will select `0.5.1` and fail.

jmank88: IMO both web3 and explorer can make a best-effort default/suggestion (the current approach is a start), but there has to be a way to override to a specific version.

jmank88: For `web3`: Even for the basic case of e.g. `pragma solidity >=1.0.0;`, do we really want to compile with `v1.0.0`? Or do we want the latest stable version (maybe `v5.x.y`) for what is likely a better result?

jmank88: > In this case it makes sense to implement the same feature in web3 to keep the same UX

I agree they should be somewhat aligned, but they are different use cases. `web3` should be trying to produce the best possible output (e.g. latest valid stable makes sense). Explorer verification is intended to reproduce a previous action exactly, and must be precise and customizable in order to do so.

jmank88: Yes, a user could modify the pragma to try to inform our decision (e.g. bump from `>=1.00` to `>=5.1.3`), but that is a misuse of what it is designed for. It expresses compatibility constraints.

r-gochain: so let me then add a way to specify a version in both explorer and web3 and leave auto-suggestion(the dumb one we have now - not sure it's a good idea to try to resolve version constraints) if nothing was selected

jmank88: I agree that resolving complex constraints could be a mess, and can be improved later instead. As long as users can make an explicit custom choice, then they aren't left behind if the suggestion is incompatible or outdated.

PR #246, opened 2019-03-27
Custom rpc, closes #241

### LINKED ISSUE #241: Wallet: Support custom RPC
It seems like wallet is hard coded to only support the public main/test rpc endpoints. Instead, it should connect to whatever rpc url the web server is using to fetch data. i.e. if you are running a local dev network, or a private network, you should be able to use explorer's wallet.

bergauz:


jmank88:
Can this be rebased on master, or wait until #242 is merged?

Issue #238, opened 2019-03-14
Fix Integration Tests


jmank88:
The go integration tests are broken (`go test -tags=integration ./...`). After they are fixed, they should be re-enabled here: https://github.com/gochain-io/explorer/blob/master/Dockerfile#L14-L15
Follow up from #237

PR #249, opened 2019-03-30
Fix Integration Tests #238

r-gochain:


run_tests.sh
```diff
@@ -16,9 +29,6 @@ echo "Docker logs for server"
...
 echo "Trying curl"
 docker run --rm --network="container:$varA" byrnedo/alpine-curl -f http://localhost:8080/
-docker run --rm --network="container:$varA" byrnedo/alpine-curl -f http://localhost:8080/api/blocks/10 
+docker run --rm --network="container:$varA" byrnedo/alpine-curl -f http://localhost:8080/api/blocks/10
 docker run --rm --network="container:$varA" byrnedo/alpine-curl -f http://localhost:8080/api/blocks/10/transactions
-for imagename in test_web3 test_mongo test_explorer_grabber test_explorer_server; do
-    echo "* Trying to stop ${imagename}..."
-    [ -z $(docker ps -a -f "name=$imagename" --format '{{.Names}}') ] || docker rm -f $imagename
-done
+cleanup_containers
```
jmank88: I added the `trap`s because this kept hanging and I had to kill it with an interrupt and clean it up manually. It's good that it will clean it self up beforehand now, but won't it still leave containers running if it's interrupted? Can we move this closing cleanup call into a `trap` up top instead? It should also make the initial call unnecessary, because it will never leave anything running in the first place.

jmank88: Less important, but IMO a per-container call is cleaner (e.g. `trap cleanup_container(test_explorer_grabber) EXIT SIGINT SIGTERM`). That way it only tries to clean what has actually been launched.

jmank88: Another alternative would be to use `docker-compose`.

r-gochain: this 'trap' thing was not working properly on my env(I'm running ubuntu 19, so in theory behaviour should be the same but it wasn't), but I'll check how errors/exits could be handled properly

jmank88: How was it failing? Do we just need a #!/bin/bash or something?

r-gochain: nope it was not working as it used to
1) that was the first issue https://unix.stackexchange.com/questions/314554/why-do-i-get-an-error-message-when-trying-to-trap-a-sigint-signal
2) It was not run in case of failure (figuring out now)

jmank88: Strange. I thought it was working fine for me on Ubuntu 18.04 and 18.10

jmank88:
I still think we should try `docker-compose` (since we're essentially re-implementing it's functionality manually) but I can fiddle with that in a follow-up PR.

r-gochain:
> I still think we should try `docker-compose` (since we're essentially re-implementing it's functionality manually) but I can fiddle with that in a follow-up PR.

yep it makes sense, but need to fix integration tests (on circle-ci)

r-gochain:
@jmank88 I tried docker-compose approach and it seems like it would introduce some hacks and be a bit more complicated (due to circle-ci features - secure/safe mounting and networking) - ie it's not possible to mount volume from the local dir (which could be solved using the same approach I used here - docker cp) and networking works in a bit weird way (I was not able to mount ports to localhost on circle, but maybe need to play a bit with flags/options)

PR #251, opened 2019-04-01
HTTP HEAD closes #217

### LINKED ISSUE #217: HTTP HEAD
HTTP HEAD requests return HTTP 405 or 404. We should support HEAD from the root and from `/api` (api should only return 200 if explorer can ping the backing DB).

r-gochain:


server/main.go
```diff
@@ -200,6 +201,7 @@ func main() {
 		})
 
 		r.Route("/", func(r chi.Router) {
+			r.Head("/", pingDB)
```
jmank88: Can we just return static 200 from this one? That way the status page can have separate statuses for web and api.

server/main.go
```diff
@@ -424,3 +426,13 @@ func getBlock(w http.ResponseWriter, r *http.Request) {
 	}
 	writeJSON(w, http.StatusOK, block)
 }
+
+func pingDB(w http.ResponseWriter, r *http.Request) {
+	err := backendInstance.PingDB()
+	if err != nil {
+		errorResponse(w, http.StatusInternalServerError, err)
+	} else {
+		writeJSON(w, http.StatusOK, "PingOK")
```
jmank88: Looks like there shouldn't be a body: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD
Technically should be headers from GET request, but we only need the 200 code for our purposes.

PR #257, opened 2019-04-06
added addr qr code, resolves #256

### LINKED ISSUE #256: QR Code
We should consider displaying account QR codes on the address and wallet pages, and serving the raw image over an api url, e.g. `https://explorer.gochain.io/api/qr/<addr>` (maybe an extension as well, if it makes sense to provide more than one of `.png`, `.svg`, etc.)

bergauz:


jmank88:
Can you post screenshot of each page? Mobile too

r-gochain:
image
image
image

r-gochain:
@bergauz also please add to opened wallet page (you added to "create new wallet" page, but not to "open existing wallet" page)

jmank88:
What about dropping the QR label and just putting the image next to the text address?

bergauz:
Screenshot 2019-04-11 at 20 38 56
@jmank88 smth like this?

jmank88:
Yeah that looks good. Does it collapse underneath the hex address when too narrow?

PR #261, opened 2019-04-09
tx receipt status, resolves #259

### LINKED ISSUE #259: Tx Receipt Status
The testnet tx `0xc1368648ee04a97c06fa5b293c8743420c79b2628fd6424616bb9f78fb6565fd` failed, but you wouldn't know if from the interface:
Screenshot from 2019-04-08 18-03-25
https://testnet-explorer.gochain.io/tx/0xc1368648ee04a97c06fa5b293c8743420c79b2628fd6424616bb9f78fb6565fd

We already check gas used from the receipt, we should check the status as well. Probably Logs too, but that could be complicated and can be done separately later.

bergauz:


server/backend/db_backend.go
```diff
@@ -474,6 +478,10 @@ func (self *MongoBackend) getTransactionByHash(transactionHash string) *models.T
 		}
 		c.GasFee = new(big.Int).Mul(gasPrice, big.NewInt(int64(receipt.GasUsed))).String()
 		c.ContractAddress = receipt.ContractAddress.String()
+		c.Status = false
+		if receipt.Status != 0 {
+			c.Status = true
+		}
```
jmank88: Can we change this to == 1 ?

server/models/transaction.go
```diff
@@ -6,6 +6,7 @@ type Transaction struct {
 	TxHash          string    `json:"tx_hash" bson:"tx_hash"`
 	To              string    `json:"to" bson:"to"`
 	From            string    `json:"from" bson:"from"`
+	Status          bool      `json:"status" bson:"status"`
```
jmank88: What about `Success` instead?

jmank88:
What about combining `Block Number` and `Processed at` and calling it 'Status`? Then the three cases are:

**Status**
[Pending]

**Status**
[Failed] block number - timestamp (x ago)

**Status**
[Success] block number - timestamp (x ago)

server/backend/db_backend.go
bergauz: @jmank88 yes, would be better

bergauz:
@jmank88 I'm not sure about pending, since grabber get's txs from block, so only 2 states are available: success, failed

jmank88:
I thought it used to be possible to look up a tx directly while it is pending and it would just show no block number. We appear to have lost that behavior though.

bergauz:
@jmank88 we can use web3 direct call to get tx info, but block time is 3 sec, that is fast, so does we need this?

jmank88:
Yes I think we still need this, but as a separate change. Block time is 5 seconds, but the delay can be up to 6 or 7 seconds even with no load (in practice, I follow turbowallet links quick enough to see not found all the time). With a high enough load, a low priced tx could be pending indefinitely.

server/models/transaction.go
bergauz: let's leave it like to make it more similar to web3 tx

jmank88: I'm not sure what you mean, but how about leaving a comment like this?
```suggestion
	Status          bool      `json:"status" bson:"status"` // True: Successful, False: Failed
```

jmank88:
Screenshots?

PR #264, opened 2019-04-15
server, grabber: add locked-accounts option

jmank88:
This PR adds a `locked-accounts` options to server and grabber for the rich list and circulating supply calculations. This removes the dependency on `eth_genesisAlloc` which was unnecessary, inefficient, and only happens to have the correct addresses anyways. Now the locked accounts can be different from the accounts in the genesis allocation, and they can change over time.

We'll have to update configurations when deploying this.
Testnet: `0x2fe70f1df222c85ad6dd24a3376eb5ac32136978`
Mainnet: `0xf75b6e2d2d69da07f2940e239e25229350f8103f`


r-gochain:
LGTM

PR #273, opened 2019-04-24
Abi improvements

bergauz:


src/app/pipes/abi-method.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+})
+
+export class AbiMethodPipe implements PipeTransform {
+
+  transform(val: ABIDefinition): string {
+    if (!val) {
+      return null;
+    }
+    const inputs = val.inputs.map(input => input.name);
+    // const outputs = val.outputs.map(output => output.name + (output.name ? ' ' : '') + output.type);
```
jmank88: I personally would like to see types also, for both inputs and outputs, but not sure if that is best for general users.

src/app/pipes/to-gwei.pipe.ts
```diff
@@ -3,7 +3,7 @@ import {PipeTransform, Pipe} from '@angular/core';
 @Pipe({
   name: 'toGwei'
 })
-export class ToGwei implements PipeTransform {
+export class ToGweiPipe implements PipeTransform {
```
jmank88: Unrelated, but shouldn't this implementation be using some sort of big number?

src/app/pipes/abi-method.pipe.ts
bergauz: @jmank88 yeah, i've tried and it's getting too long and entangled

PR #275, opened 2019-04-25
Abi

r-gochain:


server/models/contract.go
```diff
@@ -1,17 +1,31 @@
 package models
 
-import "time"
+import (
+	"github.com/gochain-io/gochain/v3/accounts/abi"
+	"time"
+)
+
+type AbiItem struct {
```
jmank88: @bergauz we should add the `omitempty` option to some of these `json` tags (probably `bson` too), since each type of item has a different subset of the fields. Might have to be careful with the bools though (maybe use `*bool`) if false is treated differently than missing on the front end.

PR #288, opened 2019-05-02
Nonce & tx page fixes

### LINKED ISSUE #113: Nonce Displaying as 򅻟
https://testnet-explorer.gochain.io/tx/0xbb2895fd2a997d4d71c0065bc1428262f2d16d4aa2606e21356decac138a75e3

screenshot from 2018-09-25 09-36-42

This appears to affect every tx

### LINKED ISSUE #283: Hide Empty Tx Contract Address Field
The `Contract Address` field should only appear if the tx created a contract.
Screenshot from 2019-04-30 09-52-51
https://testnet-explorer.gochain.io/tx/0x9b7a0581897443a0194795651f3f7e1fe6eea893d86b62d3104eb4c53c0ef1b2

bergauz:
empty nonce fix, resolves #113 
 show contract addr only for successful tx in tx page, resolves #283 
argin style fix


src/app/scenes/transaction/transaction.component.ts
```diff
@@ -93,7 +93,7 @@ export class TransactionComponent implements OnInit, OnDestroy {
         finalTx.value = tx.value;
         finalTx.gas_price = tx.gasPrice;
         finalTx.gas_limit = '' + tx.gas;
-        finalTx.nonce = '' + tx.nonce;
+        finalTx.nonce = +tx.nonce;
```
jmank88: What does this do?

src/app/scenes/transaction/transaction.component.html
```diff
@@ -41,7 +41,7 @@ <h1>Explore your block</h1>
           <p>Processed at</p>
           <b>{{tx.created_at | date: 'yyyy/MM/dd HH:mm:ss'}} ({{tx.created_at | timeAgo }})</b>
         </li>
-        <li *ngIf="tx.contract_address">
+        <li *ngIf="tx.status && tx.contract_address">
```
jmank88: Is checking the status necessary/relevant? It seems like the literal `0x0000000000000000000000000000000000000000` is coming through, and a cleaner fix would be to change the backend to send null instead of a zero value `common.Address`.

bergauz:
so I've changed Nonce string->uint64, test do not pass cause api linked to explorer.gochain.io, but locally it passes

r-gochain:
There is one more problem that nonce for old txs stored in wrong format, I'll force refill that

src/app/scenes/transaction/transaction.component.html
r-gochain: @bergauz ^

Issue #285, opened 2019-05-01
Include Token Balances on Address Page


jmank88:
We already index token transfer txs for each contract. For each account involved in those txs, we should also fetch and store the token balances (like the normal GO balance information), and display them on the address page.

Something like:
```
Tokens:
12345 TURBO
56788 GOST
```
With the symbol names linking to the contracts.

This is pretty straightforward for ERC20s. If ERC721s are too messy (UI could be complicated since they could hold a ton and we might need to page through lists or something), then they can be done later in a separate change.

We could also consider adding a `Token Transfers` tab just like `Internal Transactions` from the contract page, which includes only token transfer txs to/from this account.

bergauz:
@jmank88 what's strategy to update token balance?

jmank88:
I don't know the grabber internals well enough to be sure, but it seems like it could be part of the normal GO balance update process, or an analogous process that works in the same general way. Maybe @r-gochain has some ideas.

r-gochain:
Yeah I'm going to work on this today

r-gochain:
![image](https://user-images.githubusercontent.com/40002763/57102768-f0f3aa00-6d2c-11e9-98ac-2241bf504863.png)

r-gochain:
@bergauz could you add frontend part for this ticket? 
I've added a new API endpoint(the same as /holders) - api/address/TOKEN_OWNER_ADDRESS/owned_tokens?limit=25&skip=0

jmank88:
Sweet that was easy! Fetching the internal txs to/from an account should be pretty easy too, yeah? Just another index or two on existing data?

r-gochain:
![image](https://user-images.githubusercontent.com/40002763/57164224-1dc2c280-6dfc-11e9-89eb-b77559fffe4a.png)

r-gochain:
@bergauz added token_transactions parameter (/api/address/ADDRESS_HASH/internal_transactions?token_transactions=true&limit=25&skip=0)

r-gochain:
@bergauz have you added Token transfers too?(I remember I saw it, just to confirm)

bergauz:
@r-gochain yes

bergauz:
closed by #290, #300

PR #290, opened 2019-05-02
Include Token Balances on Address Page #285

r-gochain:


server/main.go
```diff
@@ -326,6 +327,16 @@ func getTokenHolders(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, http.StatusOK, tokenHolders)
 }
 
+func getOwnedTokens(w http.ResponseWriter, r *http.Request) {
+	contractAddress := chi.URLParam(r, "address")
+	skip, limit := parseSkipLimit(r)
+	tokens := &models.OwnedTokenList{
+		OwnedTokens: []*models.TokenHolder{},
```
jmank88: Unnecessary declaration?

jmank88:
@bergauz screenshot? I'm not sure tokens should be under balance. Maybe their own 'Tokens' section, or even a tab at the bottom? And the symbol should link to the token contract.

r-gochain:
image

r-gochain:
Yeah I think 1) separate tab (at the top) would makes sense (it could be many many tokens on the same address) 2) token_symbols should be clickable (there is contract_address in payload)

@bergauz

jmank88:
I was thinking separate tab at the bottom, as in `[ Transactions | Tokens ]`. What would a tab at the top look like?

If at the bottom, each row could contain some more information since we have more room, e.g.:
`x <symbol>, <token name>, <linked contract address>`

bergauz:
@jmank88 yeah, better to do at the bottom, i cant image tab at the top

src/app/scenes/address/address.component.html
```diff
@@ -175,7 +170,50 @@ <h1>Explore your block</h1>
...
+              <app-pagination [total]="tokensQueryParams.totalPage"
+                              (onSelect)="tokensQueryParams.toPage($event)"></app-pagination>
+            </div>
+          </ng-container>
+          <ng-template #emptyTokens>
+            <div class="empty">No tokens</div>
+          </ng-template>
+        </tab-content>
+      </tab>
+      <tab name="internal_transactions" [title]="addr.contract ? addr.token_symbol + ' Transactions' : 'Token Transactions'">
```
jmank88: Is this covering both the internal contract txs and the address token txs in the same tab? I think they need to be separate, because a contract page would have both: `<symbol> Txs` for the internal txs of that contract, and `Token Txs` for the txs internal to _other_ token contracts and which are to/from this contract (as a holder).

bergauz: @jmank88 yes, fixed

PR #298, opened 2019-05-08
preload contract abi

### LINKED ISSUE #293: Since we already know common contract types, preload the ABI text box
Rather than having to click the button first.

bergauz:
preload known contract abi, resolves #293


server/utils/constants.go
```diff
@@ -1,6 +1,21 @@
...
+	Constant        bool          `json:"constant" bson:"constant"`
+	Inputs          abi.Arguments `json:"inputs" bson:"inputs"`
+	Name            string        `json:"name" bson:"name"`
+	Outputs         abi.Arguments `json:"outputs" bson:"outputs"`
+	Payable         bool          `json:"payable" bson:"payable"`
+	StateMutability string        `json:"stateMutability" bson:"stateMutability"`
+	Type            string        `json:"type" bson:"type"`
+}
+
+type InterfaceName string
```
jmank88: What about renaming this to `FunctionName`?

bergauz: @jmank88 maybe AbiDefinition like in web3?

bergauz:
so know only front uses json, right now it is needless at server

server/utils/constants.go
jmank88: I personally think that `AbiDefinition` is a confusing/ambiguous name because it's not clear if it's a full ABI definition (collection of functions), or a single function. The whole web3 ecosystem seems split between `function`/`method`/`callable`, but I think that any of those make it more clear.

jmank88: I don't want to get too far off track from the PR, but `InterfaceData` could be renamed the same way, and its fields could use some short docs to clarify meaning.

jmank88: I like `FunctionName`/`FunctionData`, and I think the `Value` hash is normally called `ID`.

r-gochain: @bergauz figured out?

inkberk: renamed

PR #301, opened 2019-05-10
Erc20 output improvement

bergauz:
needs to be merged with rebase


src/app/modules/wallet/wallet-send/wallet-send.component.ts
```diff
@@ -283,6 +284,25 @@ export class WalletSendComponent implements OnInit {
...
+    let value: string = (<HTMLInputElement>event.target).value;
+    if (value) {
+      value = (new BigNumber(value)).multipliedBy('1e' + this.addr.decimals).toString();
+      if (/e+/.test(value)) {
+        const parts = value.split('e+');
+        let first = parts[0].replace('.', '');
+        const zeroes = parseInt(parts[1], 10) - (first.length - 1);
+        for (let i = 0; i < zeroes; i++) {
+          first += '0';
+        }
```
jmank88: Is there a way to control how the `BigNumber` is formatted, rather than picking it apart and reformatting manually afterwards?

bergauz: @jmank88 i wasn't able to bypass scientific notation, so used this path

Issue #233, opened 2019-02-27
Extra extra data


jmank88:
The 'Extra Data' field is displayed in full, but only the first 32 bytes of vanity data are human readable:
![screenshot from 2019-02-27 11-43-58](https://user-images.githubusercontent.com/1194128/53511050-0e826a00-3a85-11e9-93d3-d6ae081389e4.png)
From https://explorer.gochain.io/block/0x3a5bdb2e1e396ed846dd055d42a625d30446183f5f3f3f180d04b5e5cb99be37

The first 32 vanity bytes should be shown separately.

The rest of the data is not included on every block. We could also consider displaying the full data, but should convert it to a human readable form. The `consensus/clique` package has helper functions for extracting the relevant parts:

- [ExtraVanity](https://godoc.org/github.com/gochain-io/gochain/consensus/clique#ExtraVanity)
- [ExtraHasVote](https://godoc.org/github.com/gochain-io/gochain/consensus/clique#ExtraHasVote)
- [ExtraCandidate](https://godoc.org/github.com/gochain-io/gochain/consensus/clique#ExtraCandidate)
- [ExtraIsVoterElection](https://godoc.org/github.com/gochain-io/gochain/consensus/clique#ExtraIsVoterElection)

Additionally, the `Nonce` field of the header indicates either an auth (`0xffffffffffffffff`) or drop (`0x0000000000000000`) vote. This is the only use for `Nonce`, and it can be removed from its current place in the interface: 
![screenshot from 2019-02-27 11-59-22](https://user-images.githubusercontent.com/1194128/53511982-3b378100-3a87-11e9-95f5-60e41e491238.png)

Ideally, for a regular block we'd see only:

```
Vanity Data: xxxxxx
```

And for a block containing a vote (ExtraHasVote), we'd see:

```
Vanity Data: xxxx
Vote: Authorize/Drop `0x1234` to sign/vote
```

jmank88:
The homepage `Extra Data` column should also only be vanity data (not sure what it is currently).

r-gochain:
![image](https://user-images.githubusercontent.com/40002763/59714894-56d3bc80-921b-11e9-8ebb-e1dc04795cb5.png)
@jmank88 thoughts?

jmank88:
IMHO combining the fields would be much more clear to the users:

> **Vote**
> Add/Remove signer/voter `0x1234`

r-gochain:
@jmank88 so Add remove - it's Nonce (0 or -1)
[...]

PR #325, opened 2019-06-17
fix for failed nonce update

r-gochain:


server/backend/db_backend.go
```diff
@@ -96,7 +96,7 @@ func (self *MongoBackend) parseBlock(block *types.Block) *models.Block {
 		ParentHash: block.ParentHash().Hex(),
 		TxHash:     block.Header().TxHash.Hex(),
 		GasUsed:    strconv.Itoa(int(block.Header().GasUsed)),
-		Nonce:      uint64(block.Nonce()),
+		Nonce:      block.Nonce(),
```
jmank88: It looks like the field and the method return are both already `uint64`, so what does this fix exactly? Isn't it just removing a redundant type conversion?

r-gochain: yep just removing redundant type conversion

jmank88: How does that fix the error? Isn't the code behaving the same either way?

r-gochain: image

jmank88: `0` is not the correct value for the nonce of that block. It is `0xffffffffffffffff`, not `0x0000000000000000`.

jmank88: https://rpc.gochain.io/x/eth_getBlockByNumber/6869909

jmank88: I don't think it is valid to pass a max uint64 to javascript, unless we are handling it carefully with a big number library. But regardless, we don't even really want those values on the front end anyways since the field is only relevant during a vote, and only has two possible values.
https://github.com/gochain-io/explorer/issues/233

r-gochain: so do you want to see hex value there?

jmank88: > so do you want to see hex value there?

No. The linked issue goes in to detail about how I thought it should be displayed. The interpretation of the hex values can probably be done on the backend to avoid this whole mess, and a simpler `bool` or `*bool` can be passed to the front end.

r-gochain: so maybe we could merge this to fix the error with non-loading block and I(or Murat) will work on #233 separately?

jmank88: I don't think that this fixes the error. I would bet that the code is compiling exactly the same with or without the superfluous type conversion, and something else is going on that prevented the crash. IMO better intermediate step would be to remove the nonce from the front end entirely, since it's not useful to the users, especially if it's always 0.

r-gochain: it crashes on backend, after this fix it doesn't crash (as on screenshot above), feel free to test yourself

jmank88: > BSON has no uint64 type, and value is too large to fit correctly in an int64

Doesn't this error suggest that the nonce cannot be stored in mongo as a uint64?

jmank88: > it crashes on backend, after this fix it doesn't crash (as on screenshot above), feel free to test yourself

There must be something else going on. The conversion doesn't do anything.

r-gochain: interesting )

r-gochain: let me finish migration to gce and then I'll check 233

PR #326, opened 2019-06-19
Extra extra data #233

r-gochain:
With extra data
image

Without extra data
image

### DIFF
--- a/server/models/block.go
+++ b/server/models/block.go
@@ -18,22 +18,19 @@ type Block struct {
 	Difficulty      int64     `json:"difficulty" bson:"difficulty"`
 	TotalDifficulty int64     `json:"total_difficulty" bson:"total_difficulty"`
 	Sha3Uncles      string    `json:"sha3_uncles" bson:"sha3_uncles"`
-	ExtraData       string    `json:"extra_data" bson:"extra_data"`
+	ExtraData       string    `json:"-" bson:"extra_data"`
 	// Transactions    []string  `json:"transactions" bson:"transactions"`
 
-	ExtraAuth            bool   `json:"extra_auth" bson:"-"`
-	ExtraVanity          string `json:"extra_vanity" bson:"-"`
-	ExtraHasVote         bool   `json:"extra_has_vote" bson:"-"`
-	ExtraCandidate       string `json:"extra_candidate" bson:"-"`
-	ExtraIsVoterElection bool   `json:"extra_is_voter_election" bson:"-"`
+	Extra ExtraDataStruct `json:"extra" bson:"-"`
 }
 
 type LightBlock struct {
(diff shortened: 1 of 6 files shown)


r-gochain:
@jmank88 could you please?

server/models/block.go
```diff
@@ -12,14 +12,20 @@ type Block struct {
...
 	TotalDifficulty int64     `json:"total_difficulty" bson:"total_difficulty"`
 	Sha3Uncles      string    `json:"sha3_uncles" bson:"sha3_uncles"`
 	ExtraData       string    `json:"extra_data" bson:"extra_data"`
 	// Transactions    []string  `json:"transactions" bson:"transactions"`
+
+	ExtraAuth            bool   `json:"extra_auth" bson:"-"`
+	ExtraVanity          string `json:"extra_vanity" bson:"-"`
+	ExtraHasVote         bool   `json:"extra_has_vote" bson:"-"`
+	ExtraCandidate       string `json:"extra_candidate" bson:"-"`
+	ExtraIsVoterElection bool   `json:"extra_is_voter_election" bson:"-"`
```
jmank88: Would it be clearer to group these into a struct which is only set when `clique.ExtraHasVote` is true?
```go
type Vote struct {
 Candidate string
 Voter bool
 Auth bool
}
```

src/app/scenes/block/block.component.html
```diff
@@ -61,10 +57,21 @@ <h1>Explore your block</h1>
...
+            <p>Extra Data</p>
+            <b>{{block.extra_vanity }}</b>
+          </li>                          
+        </ul>
+      </li>
+        <ng-container *ngIf="block.extra_has_vote; else empty">
+          <br/>
+              <li>
+                <p>Vote</p>
+                <b>{{block.extra_auth ? "Add" : "Remove"}} {{block.extra_candidate}}</b> to <b>{{block.extra_is_voter_election ? "vote" : "sign"}}</b>
```
jmank88: Reordering to `Add signer 0x1234` might wrap better on mobile.

src/app/scenes/block/block.component.html
```diff
@@ -61,10 +57,21 @@ <h1>Explore your block</h1>
...
         </li>
         <li>
-          <p>Extra Data</p>
-          <b>{{block.extra_data }}</b>
-        </li>
-      </ul>
+        <ul>
+          <li>
+            <p>Extra Data</p>
+            <b>{{block.extra_vanity }}</b>
```
jmank88: Can we use `extra_vanity` on the home page block table as well? And really anywhere else `extra_data` is being used. I'm not sure there is any use for it on the front end in the raw form.

PR #332, opened 2019-07-04
handle richlist errors gracefully

r-gochain:


server/main.go
```diff
@@ -237,15 +237,24 @@ func main() {
...
-	totalSupply, _ := backendInstance.TotalSupply()
-	total := new(big.Rat).SetFrac(totalSupply, wei) // return in GO instead of wei
-	w.Write([]byte(total.FloatString(18)))
+	totalSupply, err := backendInstance.TotalSupply()
+	if err != nil {
+		total := new(big.Rat).SetFrac(totalSupply, wei) // return in GO instead of wei
+		w.Write([]byte(total.FloatString(18)))
+	} else {
+		writeJSON(w, http.StatusInternalServerError, err)
+	}
```
jmank88: The error check is flipped on this one and circulating.

jmank88: I tend to prefer:
```go
if err != nil {
  // handle
  return
}
// normal path
```
by default, unless there is a good reason not to.

server/main.go
```diff
@@ -268,16 +277,24 @@ func getCurrentStats(w http.ResponseWriter, r *http.Request) {
...
+			writeJSON(w, http.StatusInternalServerError, err)
+		} else {
+			bl := &models.Richlist{
+				Rankings:          []*models.Address{},
+				TotalSupply:       new(big.Rat).SetFrac(totalSupply, wei).FloatString(18),
+				CirculatingSupply: new(big.Rat).SetFrac(circulatingSupply, wei).FloatString(18),
+			}
+			bl.Rankings = backendInstance.GetRichlist(skip, limit)
+			writeJSON(w, http.StatusOK, bl)
+		}
```
jmank88: In cases like this, some `return`s avoid the nested `else`s:
```go
if err != nil {
	writeJSON(w, http.StatusInternalServerError, err)
	return
} 
skip, limit := parseSkipLimit(r)
circulatingSupply, err := backendInstance.CirculatingSupply()
if err != nil {
	writeJSON(w, http.StatusInternalServerError, err)
	return
} 
bl := &models.Richlist{
	Rankings:          []*models.Address{},
	TotalSupply:       new(big.Rat).SetFrac(totalSupply, wei).FloatString(18),
	CirculatingSupply: new(big.Rat).SetFrac(circulatingSupply, wei).FloatString(18),
}
bl.Rankings = backendInstance.GetRichlist(skip, limit)
writeJSON(w, http.StatusOK, bl)	
```

server/main.go
```diff
@@ -237,15 +237,24 @@ func main() {
 }
 
 func getTotalSupply(w http.ResponseWriter, r *http.Request) {
-	totalSupply, _ := backendInstance.TotalSupply()
-	total := new(big.Rat).SetFrac(totalSupply, wei) // return in GO instead of wei
-	w.Write([]byte(total.FloatString(18)))
+	totalSupply, err := backendInstance.TotalSupply()
+	if err != nil {
```
jmank88: These checks are backwards. Could do this, or flip bodies and drop the else.
```suggestion
	if err == nil {
```

server/main.go
```diff
@@ -237,15 +237,24 @@ func main() {
...
+		writeJSON(w, http.StatusInternalServerError, err)
+	}
 }
 
 func getCirculating(w http.ResponseWriter, r *http.Request) {
-	circulatingSupply, _ := backendInstance.CirculatingSupply()
-	circulating := new(big.Rat).SetFrac(circulatingSupply, wei) // return in GO instead of wei
-	w.Write([]byte(circulating.FloatString(18)))
+	circulatingSupply, err := backendInstance.CirculatingSupply()
+	if err != nil {
```
jmank88: Same here.
```suggestion
	if err == nil {
```

dome:
@r-gochain  richlist work fine but balance from api show only 0 
V0.4.83 correct
https://exp.tch.in.th/api/address/0xb82fF5c8C8b5D1E72B9C1e6875Bf2fE8293202a0
V0.4.84 show 0
http://dc5.dc0.us:8080/api/address/0xb82fF5c8C8b5D1E72B9C1e6875Bf2fE8293202a0

r-gochain:
@dome it would be great if you open separate ticket for that, but I've just checked and didn't find any issues, could you check to which RPC endpoint your exporer instance is pointed also you could try to get the balance directly from rpc node

curl https://rpc.tch.in.th/ -H 'content-type: application/json;' --data-binary '[{"id":"94455d269e0ee68abf391d9923d26b13","jsonrpc":"2.0","method":"eth_getBalance", "params":["0xb82fF5c8C8b5D1E72B9C1e6875Bf2fE8293202a0","pending"]}]'

PR #333, opened 2019-07-04
refactored get address, now force update address when it's not found

r-gochain:


server/backend/backend.go
```diff
@@ -118,7 +118,18 @@ func (self *Backend) GetRichlist(skip, limit int) []*models.Address {
 
 }
 func (self *Backend) GetAddressByHash(hash string) *models.Address {
-	return self.mongo.getAddressByHash(common.HexToAddress(hash).Hex())
+	addressHash := common.HexToAddress(hash).Hex()
+	address := self.mongo.getAddressByHash(addressHash)
+	balance, err := self.BalanceAt(addressHash, "latest")
```
jmank88: Should the error be logged at least? Or returned, since `getAddress` is just returning an empty response now.

Issue #334, opened 2019-07-10
Signers Page


jmank88:
Add a new page for signer information at `/signers` and linked as "Signers" in the main nav. This page should contain a table of signer account information (address, name, url, region) with stats about blocks signed (count, %), and a pie chart like the following (from https://www.blockchain.com/en/pools):

![Screenshot from 2019-07-08 11-29-44](https://user-images.githubusercontent.com/1194128/60998286-37154d00-a31e-11e9-9340-4c73f6143164.png)

Caption: `Block Signers for 9768 blocks from <start ts> to <end ts>`

| Account | Name | Region | Blocks | % |
| ------ | --- | --- | --- | --- |
| [0x1](https://explorer.gochain.io/addr/0x1) | [GoChain 1](https://gochain.io) | North America | 100 | 4% |
| [0x2](https://explorer.gochain.io/addr/0x2) | [GoChain 2](https://gochain.io) | South America | 102 | 4.1% |

The data should be available for a fixed set of ranges, like 1d, 7d, 30d, and groupable by region if possible.

One possible data format:

```go
type SignerDistribution struct {
  Start, End uint64 // unix ts for queried ranged
  TotalBlocks uint64 
  Signed []struct {
    Address common.Address
    Blocks uint64
  }
}
```

The service would then return a fixed set of ranges (e.g. 1d, 7d, 30d), which can each be independently cached and refreshed on their own schedules.

```go
type SignerDistributions struct {
  Day, Week, Month SignerDistribution  // 1d, 7d, 30d
  Signers []struct {
    Address common.Address
    Name string
    URL string
    Region string
  }
}
```

The extra `Signers` information will have to be loaded from a configuration file. This info can be used to enhance the UI in other places, e.g. by adding name/url to pages for accounts, blocks, rich list (related #125), and the homepage. Example `signers.json`:
```json
[
  {
    "address": "0xA",
    "name": "GoChain 1",
    "url": "https://gochain.io",
    "region": "North America"
  },
  ...
]
```

bergauz:
closed by #336

PR #336, opened 2019-07-15
signers stats

r-gochain:
draft
#334


server/models/config.go
```diff
@@ -0,0 +1,27 @@
+package models
+
+import (
+	"encoding/json"
+
+	"github.com/gochain-io/gochain/v3/common"
+)
+
+type Node struct {
+	Address common.Address
```
jmank88: Is this field used? Or is it a duplicate since it is also the map key?

server/models/signers_stats.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/gochain-io/gochain/v3/common"
+)
+
+type SignerStats struct {
+	Name        string
+	URL         string
+	Region      string
+	Signer      common.Address `json:"signer"`
+	BlocksCount int            `json:"blocks_count"`
```
jmank88: I was thinking that these Node fields could be separate so that they aren't duplicated here, and because we'll want them to be available system wide to use for other pages. Since it's static, it could be loaded once and kept on hand. Then this struct can be dropped, and just be a `map[common.Address]uint64` instead.

server/backend/db_backend.go
```diff
@@ -649,6 +654,53 @@ func (self *MongoBackend) getStats() *models.Stats {
 	}
 	return s
 }
+func (self *MongoBackend) getSignerStatsForRange(rangeDays int, nodes map[common.Address]models.Node) []models.SignerStats {
```
jmank88: This would be more testable if it accepted a `time.Time` and a `time.Duration`. It would also allow the call to `getBlockRange()` to use the exact same range.
```go
getSignerStatsForRange(end time.Time, dur time.Duration, ...
getBlockRange(end time.Time, dur time.Duration, ...
```

settings.json
```diff
@@ -0,0 +1,122 @@
+[
```
jmank88: Should this file live in our deployment configurations?

server/backend/db_backend.go
```diff
@@ -649,6 +654,54 @@ func (self *MongoBackend) getStats() *models.Stats {
...
+		log.Info().Err(err).Msg("Cannot get block number")
+	}
+	resp.StartBlock = startBlock.Number
+	resp.EndBlock = endBlock.Number
+	return resp
+}
+
+func (self *MongoBackend) getSignersStats(nodes map[common.Address]models.Node) []models.SignersStats {
+	var stats []models.SignersStats
+	kvs := map[string]time.Duration{"daily": -24 * time.Hour, "weekly": -7 * 24 * time.Hour, "monthly": -30 * 24 * time.Hour}
```
jmank88:
```suggestion
	const day = -24 * time.Hour
	kvs := map[string]time.Duration{"daily": day, "weekly": 7 * day, "monthly": 30 * day}
```

settings.json
r-gochain: Left as an example(and for CI will open additional PR to deployments configs

bergauz:
screencapture-localhost-4200-signers-2019-07-19-19_33_21

Dockerfile
```diff
@@ -23,6 +23,7 @@ WORKDIR /explorer
 RUN apk add --no-cache ca-certificates docker
 COPY --from=backend_builder /tmp/gochain/* /usr/local/bin/
 COPY --from=frontend_builder /explorer/dist/* /explorer/
+COPY settings.json /explorer/
```
jmank88: We shouldn't bundle this file because it is mainnet specific, and won't apply to any other networks.

server/main.go
```diff
@@ -126,6 +128,12 @@ func main() {
 			Usage:       "mongo database name",
 			Destination: &dbName,
 		},
+		cli.StringFlag{
+			Name:        "config-file, config",
+			Value:       "./settings.json",
```
jmank88: There shouldn't be a default value, and this file should be entirely optional - everything else should still work when the file is missing or empty.

Would be good to support an environment variable too.

If we want this to be a generic settings/config file, then we should change the format since it only has signers right now. Otherwise, this flag name should be something like `signers`.

server/main.go
```diff
@@ -164,7 +172,17 @@ func main() {
 			lockedAccounts[i] = common.HexToAddress(l).Hex()
 		}
 
-		backendInstance = backend.NewBackend(mongoUrl, rpcUrl, dbName, lockedAccounts)
+		data, err := ioutil.ReadFile(configFile)
```
jmank88: This whole block should only be attempted if the value is set, leaving `nodes` nil otherwise (or maybe an empty map if necessary).
```go
var nodes map[common.Address]models.Node
if configFile != "" {
  ...
}
```

server/models/signers_stats.go
jmank88: The _code_ is a little less duplicated this way, but I was talking about _data_ duplication - each set of stats contains all of the signers' `Node`s, and this information is sent over the wire every time, even though it is completely static. Additionally, other pages in the system will need this same data, so it would be much more efficient to send it once and save it in the front end to have on hand for any page that might contain an address.

server/backend/backend.go
```diff
@@ -25,6 +25,7 @@ type Backend struct {
 	dockerhubAPI          *DockerHubAPI
 	reCaptchaSecret       string
 	lockedAccounts        []string
+	nodes                 map[common.Address]models.Node
```
jmank88: `signers`?

jmank88:
@bergauz Looks pretty good. I think the chart could be centered. Swapping out the tabs for centered radio pills or something might look nice too (and would get rid of that horizontal separator that doesn't really add value here).
Screenshot from 2019-07-19 10-48-15

jmank88:
Can we add a caption in between the chart and the table with the total block and range information? Options:
```
Block signers for 9768 blocks from <start ts> to <end ts>.
```
```
Block signers for 9768 blocks from #1234 (<start ts>) to #3456 (<end ts>).
```

Dockerfile
r-gochain: Do you think in this case we should completely hide that link in UI?

server/main.go
r-gochain: Yeah as we discussed separately I'll rename to signers

server/backend/backend.go
r-gochain: Yeah I'll rename

server/models/signers_stats.go
r-gochain: Caching on frontend makes sense, I'll discuss with @bergauz

Dockerfile
jmank88: No, it's still useful statistics, we just will have bare addresses without name/url/region.

r-gochain: Ya already figured out this, makes sense to me, need just to show it properly on frontend

bergauz:
@jmank88 added caption, centered chart, sortable tables and tabs gonna take time, can we do it in a new issue?

bergauz:
there's default sort by block qty

jmank88:
> can we do it in a new issue?

I think it would be better to do it all at once before merging, but if we need to see it live on testnet first then that is fine - just wait until it's done to push to mainnet.

r-gochain:
image

r-gochain:
Let's test on testnet first (together with the new design)

src/assets/styles/scss/_card.scss
```diff
@@ -10,7 +10,7 @@
   word-wrap: break-word;
   background-color: $card-bg;
   background-clip: border-box;
-  border: $card-border-width solid $card-border-color;
+  /*border: $card-border-width solid $card-border-color;*/
```
jmank88: Why is this commented out? Can we remove it, or leave a note about why?

PR #337, opened 2019-07-15
new responsive design

bergauz:


treeder:
Let's get this on testnet so we can see it.

r-gochain:
@treeder sure, I'll deploy after some testing

PR #343, opened 2019-08-20
added range block limit

r-gochain:


grabber/main.go
```diff
@@ -54,6 +55,12 @@ func main() {
 			Usage:       "refill from this block",
 			Destination: &startFrom,
 		},
+		cli.Uint64Flag{
+			Name:        "block-range-limit, b",
+			Value:       9999,
```
jmank88: The limit is actually enforced as an inclusive maximum, so you can use nicer numbers.
```suggestion
			Value:       10000,
```

PR #366, opened 2019-09-17
Wallet improvement

bergauz:
Screenshot 2019-09-17 at 18 06 40


r-gochain:
I believe private key should be hidden and available only on click or on copy

r-gochain:
image

r-gochain:
@bergauz ?

PR #376, opened 2019-09-20
refactor backfill

jmank88:
This PR proposes some refactoring to the grabber backfill process, and some docs and renaming.

### DIFF
--- a/grabber/main.go
+++ b/grabber/main.go
@@ -8,10 +8,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/gochain-io/explorer/server/models"
 	"go.uber.org/zap"
 
 	"github.com/gochain-io/explorer/server/backend"
+	"github.com/gochain-io/explorer/server/models"
+
 	"github.com/gochain-io/gochain/v3/common"
 
 	"github.com/blendle/zapdriver"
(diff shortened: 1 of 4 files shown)


grabber/main.go
```diff
@@ -142,72 +142,96 @@ func getLatestBlockNumber(importer *backend.Backend) int64 {
...
 }
+
+// backfill continuously loops over all blocks in reverse order, verifying that all
+// data is loaded and consistent, and reloading as necessary.
 func backfill(importer *backend.Backend, startFrom int64) {
 	blockNumber := startFrom
 	if startFrom <= 0 {
 		blockNumber = getLatestBlockNumber(importer)
 	}
+	var hash common.Hash // Expected hash.
```
jmank88: Tracking the parent hash along the way allows dropping the call to `checkParentForBlock()`, which should also eliminate some redundant calls.

grabber/main.go
```diff
@@ -142,72 +142,96 @@ func getLatestBlockNumber(importer *backend.Backend) int64 {
...
+			// No expected hash for latest.
+			hash = common.Hash{}
 		}
 	}
 }
 
-func checkParentForBlock(importer *backend.Backend, blockNumber int64, numBlocksToCheck int) {
-	numBlocksToCheck--
+// checkAncestors scans the ancestors of a block to verify that they exists and their hashes match.
+func checkAncestors(importer *backend.Backend, blockNumber int64, ancestorsToCheck int) {
```
jmank88: I found the recursion a little difficult to read (and the error case could infinitely retry/recurse!), so I rewrote it in to a loop, and changed the error to be fatal, which is in line with how it is handled elsewhere (retry happens internally).

Issue #286, opened 2019-05-01
List of all ERC20 and ERC721 contracts?


treeder:
Would it make sense to have a list of all these somewhere?

jmank88:
I was thinking the same thing after creating #285.

r-gochain:
What kind of information put there? just links to contracts or amount of transactions/amount of holders/owner/total supply?

@treeder @jmank88

treeder:
Just name and symbol with link is probably fine. We may want to filter them out based on activity or number of holders or something so we don't list a bunch of test contracts. 

Saying that, let's hold off on this one for a bit actually, it may not be all that useful... And it may make people apprehensive about deploying their contract if we make it too easy to find it before they want people to find it.

treeder:
That could work.

PR #379, opened 2019-09-23
list of contracts, related to #286

bergauz:
Screenshot 2019-09-24 at 04 14 29

### DIFF
--- a/server/backend/db_backend.go
+++ b/server/backend/db_backend.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/big"
+	"regexp"
 	"strconv"
 	"time"
 
@@ -611,10 +612,10 @@ func (self *MongoBackend) getContracts(filter *models.ContractsFilter) []*models
 	var sortQuery string
 	findQuery := bson.M{"contract": true}
 	if filter.TokenName != "" {
-		findQuery["token_name"] = bson.RegEx{filter.TokenName, "i"}
+		findQuery["token_name"] = bson.RegEx{regexp.QuoteMeta(filter.TokenName), "i"}
 	}
 	if filter.TokenSymbol != "" {
-		findQuery["token_symbol"] = bson.RegEx{filter.TokenSymbol, "i"}
+		findQuery["token_symbol"] = bson.RegEx{regexp.QuoteMeta(filter.TokenSymbol), "i"}
 	}
 	if filter.ErcType != "" {
 		println(filter.ErcType)
--- a/server/backend/db_backend.go
+++ b/server/backend/db_backend.go
@@ -637,7 +637,6 @@ func (self *MongoBackend) getContracts(filter *models.ContractsFilter) []*models
 		findQuery["token_symbol"] = bson.RegEx{regexp.QuoteMeta(filter.TokenSymbol), "i"}
 	}
 	if filter.ErcType != "" {
-		println(filter.ErcType)
 		findQuery["erc_types"] = filter.ErcType
 	}
 	if filter.SortBy != "" {
--- a/server/backend/db_backend.go
+++ b/server/backend/db_backend.go
@@ -21,6 +21,9 @@ import (
 
 var wei = big.NewInt(1000000000000000000)
 
+const defaultFetchLimit = 100
+const defaultSkip = 0
+
 type MongoBackend struct {
 	host         string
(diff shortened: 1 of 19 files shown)


jmank88:
Awesome!

What does it look like on mobile?

I wonder how this column ordering would look:
` # Symbol Name Type Txs Internal-txs Holders Token-txs Address`

That row 3 is already pretty cramped with 5 interfaces - we should plan for handling even more smoothly. Since the `Type` column isn't a single sortable value like the others, what if it wasn't a column, and the types were listed underneath the other info from left to right? Looking at the code, I see this is a table, so maybe not a feasible tweak right now.

server/backend/db_backend.go
```diff
@@ -603,6 +606,47 @@ func (self *MongoBackend) updateContract(contract *models.Contract) bool {
 	return true
 }
 
+func (self *MongoBackend) getContracts(filter *models.ContractsFilter) []*models.Address {
+	var addresses []*models.Address
+	var sortQuery string
+	findQuery := bson.M{"contract": true}
+	if filter.TokenName != "" {
+		findQuery["token_name"] = bson.RegEx{filter.TokenName, "i"}
```
jmank88: Are we allowing users to submit arbitrary regex to the database? That seems exploitable.

bergauz:
Screenshot 2019-09-25 at 00 14 07
Screenshot 2019-09-25 at 00 14 22

bergauz:
about table if columns gonna increase we can replace with icons like in gochain stats

jmank88:
That looks good. I still think the ordering can be improved, one of these would make sense in terms of importance to the user, and alignment in the table columns:
` # Symbol Name Type Txs Internal-txs Holders Token-txs Address`

bergauz:
Screenshot 2019-09-26 at 14 47 58

jmank88:
Looks good! Needs some rebase/merge cleanup though it looks like.

server/backend/db_backend.go
```diff
@@ -605,6 +609,47 @@ func (self *MongoBackend) updateContract(contract *models.Contract) bool {
...
+	}
+	err := self.mongo.
+		C("Addresses").
+		Find(findQuery).
+		Sort(sortQuery).
+		Skip(filter.Skip).
+		Limit(filter.Limit).
+		All(&addresses)
+	if err != nil {
+		log.Debug().Err(err).Msg("GetContracts")
```
jmank88: This is the trouble. Need to pass in the new logger type, and this is probably ERROR level, isn't it?
```suggestion
		lgr.Error("Failed to query contracts", zap.Error(err))
```

bergauz: @jmank88  yes, fixed

server/models/address.go
```diff
@@ -27,5 +27,5 @@ type Address struct {
 }
 
 type AddressesList struct {
-	Adresses []*Address `json:"adresses"`
+	Addresses []*Address `json:"addresses"`
```
jmank88: I checked to see if this was in the public API, but there aren't actually any usages anywhere. Can this struct be removed?

server/backend/db_backend.go
```diff
@@ -605,6 +609,47 @@ func (self *MongoBackend) updateContract(contract *models.Contract) bool {
...
+	}
+	err := self.mongo.
+		C("Addresses").
+		Find(findQuery).
+		Sort(sortQuery).
+		Skip(filter.Skip).
+		Limit(filter.Limit).
+		All(&addresses)
+	if err != nil {
+		self.Lgr.Debug("GetContracts")
```
jmank88: Isn't this error important? Also, this logging library includes `caller: <package>/<file>:<linenumber>` automatically, so descriptive messages are more helpful:
```suggestion
		self.Lgr.Error("Failed to query contracts", zap.Error(err))
```

r-gochain:
@bergauz could you resolve the conflicts?

server/backend/db_backend.go
```diff
@@ -622,6 +626,47 @@ func (self *MongoBackend) updateContract(contract *models.Contract) error {
 	return nil
 }
 
+func (self *MongoBackend) getContracts(filter *models.ContractsFilter) []*models.Address {
+	var addresses []*models.Address
+	var sortQuery string
+	findQuery := bson.M{"contract": true}
```
r-gochain: add index {contract:1}

PR #381, opened 2019-09-24
Zapdriver  #378

### LINKED ISSUE #378: Switch to zap logger

r-gochain:
migrating to zap/zapdriver

closes  #378


grabber/main.go
```diff
@@ -98,7 +98,7 @@ func main() {
 	}
 	err := app.Run(os.Args)
 	if err != nil {
-		log.Fatal().Err(err).Msg("Run")
+		fmt.Fprintf(os.Stderr, "Failed to start app: %v\n", err)
```
jmank88: How about just `Fatal error`, since this could happen at any point, not just after starting.
This should also go through the logger - b/c anything that doesn't risks silently being dropped, or at least overlooked by not being structured correctly (e.g. we lose severity).

server/backend/backend_test.go
```diff
@@ -3,25 +3,38 @@
...
+func TestMain(m *testing.M) {
+	lgr, err := zapdriver.NewDevelopment()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to create logger: %v\n", err)
+		os.Exit(1)
+	}
+	testBackend = NewBackend("127.0.0.1:27017", "https://rpc.gochain.io", "testdb", nil, nil, lgr)
+	code := m.Run()
+	os.Exit(code)
+}
```
jmank88: zap has `package zaptest` for test loggers: `zaptest.NewLogger(t)`

Using it will require having one `*Backend` per test, instead of a global var, but that is more idiomatic and safer anyways.

server/backend/backend_test.go
```diff
@@ -3,25 +3,38 @@
...
 	var block types.Block
 	if err := rlp.DecodeBytes(blockEnc, &block); err != nil {
-		log.Fatal().Err(err).Msg("decode error")
+		fmt.Println(err)
```
jmank88: Shouldn't this still be fatal?

server/recaptcha.go
```diff
@@ -24,7 +23,6 @@ func verifyReCaptcha(token, reCaptchaSecret, action, remoteIP string) error {
 	}
 	resp, err := http.PostForm(RECAPTCHA_URL, params)
 	if err != nil {
```
jmank88: Should this still be logged?

server/recaptcha.go
```diff
@@ -24,7 +24,7 @@ func verifyReCaptcha(token, reCaptchaSecret, action, remoteIP string) error {
...
-		log.Fatal().Err(err).Msg("error occurred during making recaptcha request")
+		logger.Fatal("error occurred during making recaptcha request", zap.Error(err))
```
jmank88: I was thinking that you removed it because it shouldn't be _fatal_, which I think I agree with, but it still seems loggable, maybe as `Warn`? Does that make sense? `Fatal` seems extreme - since the error isn't necessarily our fault or in our control.

r-gochain: Yep (actually makes sense to remove recaptcha since it's not used)

PR #382, opened 2019-09-25
improve log messages; reduce fatals

jmank88:
This PR improves some log messages and reduces the number of fatal calls by adding retries and graceful failures.

~A 'retry' in this case is generally just 'try again on next trigger' or 'try again in 5 seconds'. Ideally we'd want something a bit more complex, like exponential backoff, but this is an incremental step forward from all the fatals before, and can be revisited in the future.~
Edit: I just found some generic retry code already in use, so I'll swap that in for some of these.

### DIFF
--- a/grabber/main.go
+++ b/grabber/main.go
@@ -8,15 +8,13 @@ import (
 	"sync"
 	"time"
 
-	"go.uber.org/zap"
-
 	"github.com/gochain-io/explorer/server/backend"
 	"github.com/gochain-io/explorer/server/models"
 
-	"github.com/gochain-io/gochain/v3/common"
-
 	"github.com/blendle/zapdriver"
+	"github.com/gochain-io/gochain/v3/common"
 	"github.com/urfave/cli"
+	"go.uber.org/zap"
 )
 
 func main() {
@@ -88,7 +86,10 @@ func main() {
 			// Ensure canonical form, since queries are case-sensitive.
 			lockedAccounts[i] = common.HexToAddress(l).Hex()
 		}
-		importer := backend.NewBackend(mongoUrl, rpcUrl, dbName, lockedAccounts, nil, logger)
+		importer, err := backend.NewBackend(mongoUrl, rpcUrl, dbName, lockedAccounts, nil, logger)
+		if err != nil {
+			return fmt.Errorf("failed to create backend: %v", err)
+		}
 		go listener(importer)
 		go updateStats(importer)
 		go backfill(importer, startFrom)
@@ -98,8 +99,9 @@ func main() {
 	}
 	err = app.Run(os.Args)
 	if err != nil {
-		logger.Fatal("Failed to start app", zap.Error(err))
+		logger.Fatal("Fatal error", zap.Error(err))
 	}
+	logger.Info("Stopping")
 }
 
 func appendIfMissing(slice []string, i string) []string {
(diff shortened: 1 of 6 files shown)


grabber/main.go
```diff
@@ -117,44 +119,50 @@ func listener(importer *backend.Backend) {
...
-			latestBlocknumber := getLatestBlockNumber(importer)
-			importer.Lgr.Debug("Getting block in listener", zap.Int64("Block Number", latestBlocknumber))
-			if prevHeader != latestBlocknumber {
-				importer.Lgr.Info("Getting block in listener", zap.Int64("Block Number", latestBlocknumber))
-				block, err := importer.BlockByNumber(latestBlocknumber)
-				if block != nil {
+			latest, err := importer.GetLatestBlockNumber()
+			if err != nil {
+				importer.Lgr.Error("Listener: Failed to get latest block number", zap.Error(err))
+				continue
```
jmank88: All the fatals that were here retry on the next tick now, so the other processes aren't disrupted.

grabber/main.go
```diff
@@ -117,44 +119,50 @@ func listener(importer *backend.Backend) {
...
 	for {
+		if blockNumber < 0 {
+			// Start over from latest.
+			hash = common.Hash{}
+			var err error
+			blockNumber, err = importer.GetLatestBlockNumber()
+			if err != nil {
+				importer.Lgr.Error("Backfill: Failed to get latest block number", zap.Error(err))
+				time.Sleep(5 * time.Second)
+				continue
```
jmank88: These fatals now continue (w/o incrementing the block number) to retry after waiting for 5 seconds. This means the process will hang on a problem block, instead of shutting everything down and restarting from the latest again. I think that this change is OK, b/c this sort of error is either something temporary that will go through eventually, or something critical that we will need to intervene on.

On the other hand, while the process is 'stuck' on this block, there could be new gaps forming near the 'latest' end of the chain, which it would have been filling in if it had been restarting from latest each time. We could restore that old behavior by jumping to latest on failure, or we could do something more sophisticated like retry a few times, then give up and loop back to latest.

grabber/main.go
```diff
@@ -208,77 +217,103 @@ func checkAncestors(importer *backend.Backend, blockNumber int64, generations in
...
-		for _, address := range addresses {
-			jobs <- address
+		select {
+		case <-t.C:
+			start := time.Now()
+			currentTime := time.Now()
+			currentBlock, err := importer.GetLatestBlockNumber()
+			if err != nil {
+				importer.Lgr.Error("Update Addresses: Failed to get latest block number", zap.Error(err))
+				continue
```
jmank88: Some of the fatals here will end the execution, which is effectively 'retried' on the next tick.

grabber/main.go
```diff
@@ -208,77 +217,103 @@ func checkAncestors(importer *backend.Backend, blockNumber int64, generations in
...
-		updateAddress(address, currentBlock, blockRangeLimit, importer)
+		err := updateAddress(address, currentBlock, blockRangeLimit, importer)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to update address %s: %v", address.Address, err))
+		} else {
+			updated++
+		}
+	}
+	if len(errs) > 0 {
+		lgr.Error("Update Addresses: worker done", zap.Int("updated", updated), zap.Int("failed", len(errs)), zap.Errors("errors", errs))
```
jmank88: Other fatals are just logged as failed attempts per address, and they are retried each time.

server/backend/db_backend.go
```diff
@@ -465,7 +474,10 @@ func (self *MongoBackend) isContract(address string) bool {
 	var c models.Address
 	err := self.mongo.C("Addresses").Find(bson.M{"address": address}).Select(bson.M{"contract": 1}).One(&c)
 	if err != nil {
-		self.Lgr.Debug("isContract", zap.String("Address", address), zap.Error(err))
+		if err == mgo.ErrNotFound {
+			return false
+		}
+		self.Lgr.Error("Failed to check if contract", zap.String("address", address), zap.Error(err))
```
jmank88: Many of these debug calls seem worth always logging, as long as they are not just 'not found' errors.

PR #384, opened 2019-09-26
use ctx with signal interrupt

jmank88:
This PR adds signal interrupt cancellation and wires `context.Context` around so that processes shutdown swiftly and gracefully. Specifically, there are no more blocking `time.Sleep` or HTTP RPC requests.

### DIFF
--- a/server/backend/eth_backend.go
+++ b/server/backend/eth_backend.go
@@ -2,6 +2,7 @@ package backend
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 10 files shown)


server/backend/eth_backend.go
```diff
@@ -46,22 +47,20 @@ type ethRequest struct {
...
-		Lgr:    lgr,
+		url: url,
+		Lgr: lgr,
 	}
 
 	return rpc
 }
 
-func (rpc *EthRPC) Call(method string, params ...interface{}) (json.RawMessage, error) {
+func (rpc *EthRPC) Call(ctx context.Context, method string, params ...interface{}) (json.RawMessage, error) {
```
jmank88: @rkononov Is this custom client necessary for some reason? I would think that we could just use `goclient.Client` or `web3.Client`, or at least leverage `rpc.Client` for the internals.

r-gochain: AFAIR we used that because original client has lack of some calls (total_supply and some others).
But if we already have them in the client we could use it, that would simplify the code a lot

jmank88: They did a good job of separating responsibilities in to a low level `rpc.Client` used by the high level `ethclient/goclient.Client`, so we could just use the `rpc.Client` directly all the time in place of some of this code, or we could use a higher level client when possible, and still make our own custom calls through the same shared underlying`rpc.Client`.

PR #389, opened 2019-09-28
bugfix for failing grabber, added tx count checker

r-gochain:


grabber/main.go
```diff
@@ -51,6 +52,11 @@ func main() {
 			Usage:       "mongo connection url",
 			Destination: &mongoUrl,
 		},
+		cli.BoolFlag{
+			Name:        "tx-count, tx",
+			Usage:       "check a transactions count for every block(a heavy operation)",
```
jmank88: Is it really that heavy? It would be nice to have this just be the default.

r-gochain: Not so heavy but much slower (call to external api instead of db operation)

PR #408, opened 2019-10-08
refactor contract scanning

jmank88:
This PR moves the contract byte code scanning in to `package utils` and adds tests. It also adds an alternate scan method with benchmarks, but it's not a better choice yet (not always faster), so it is left in the test file only for now.


server/utils/scan_test.go
```diff
@@ -0,0 +1,189 @@
...
...11628f55a4df523b3ef908290a45050565b600780546000838152600860205260408120829055600182018355919091527fa66cc928b5edb82af9bd49922954155ab7b0942694bea4ce44661d9a8736c6880155565b600160a060020a038116151561118d57600080fd5b6111978282610cad565b156111a157600080fd5b600160a060020a0316600090815260209190915260409020805460ff19166001179055565b600160a060020a03811615156111db57600080fd5b6111e58282610cad565b15156111f057600080fd5b600160a060020a0316600090815260209190915260409020805460ff19169055565b6000903b1190565b600081815260026020526040902054600160a060020a03161561097a576000908152600260205260409020805473ffffffffffffffffffffffffffffffffffffffff19169055565b805461127590600163ffffffff61128216565b9055565b80546001019055565b60008282111561129157600080fd5b50900390565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f106112d857805160ff1916838001178555611305565b82800160010185558215611305579182015b828111156113055782518255916020019190600101906112ea565b50611311929150611335565b5090565b8154818355818111156107c3576000838152602090206107c39181019083015b6106b291905b80821115611311576000815560010161133b56fea165627a7a72305820ed506b4286e03297c7446b2b7533ca9b251c17714e8a648c28a6278088c40ab70029"
```
jmank88: These are real contracts found on testnet.

server/utils/scan_test.go
```diff
@@ -0,0 +1,189 @@
...
+func (ci *contractInfo) scanAlternate(byteCode string) {
+	if len(byteCode) < 8 {
+		return
+	}
+	ci.funcs = map[EVMFunction]struct{}{}
+	for i := 0; (i + 7) < len(byteCode); i++ {
+		id := byteCode[i : i+8]
+		if fn, ok := EVMFunctionsByID[id]; ok {
+			ci.funcs[fn] = struct{}{}
+		}
```
jmank88: I thought this would do better, since it's `O(N)` and the existing is `O(NxM)`, but the benchmark cases didn't show it.

jmank88: After the constant overheads stop dominating runtime, doubling the number of functions should double the runtime of the existing, but not affect the runtime of this version.

PR #409, opened 2019-10-08
enable full abi from interact

jmank88:
This PR proposes adding `AllFunctions` and `Upgradeable` options to the contract interaction templates (and fills in a few missing functions). It would be neat to have an additional option dynamically generated with functions recognized by the bytecode scan, but we can do that later, and this will still be useful if/when we do.


jmank88:
I added `Upgradeable` too. None of the others are super common, and can probably wait for something dynamic instead.

Issue #282, opened 2019-04-29
Tx Logs


jmank88:
Tx receipts are currently fetched and used for a few fields and statuses. We should also display any receipt logs generated by the tx underneath the input data. This can be just the raw bytes as hex to start, but should eventually be parsed and displayed as solidity events, like in #281.

r-gochain:
![image](https://user-images.githubusercontent.com/40002763/58909709-74356080-871c-11e9-95fc-08a78ba138f7.png)

jmank88:
I would like to keep this open until we support parsing the events too.

jmank88:
Initial version #410 
![Screenshot from 2019-10-14 10-18-25](https://user-images.githubusercontent.com/1194128/66762619-fb81a600-ee6b-11e9-8bde-006617b113e9.png)

PR #410, opened 2019-10-09
front: embed event ABIs

jmank88:
This PR is a WIP to support event parsing in the front end (#282). This first commit embeds some event ABIs, to be used here:
Screenshot from 2019-10-09 07-33-24
(https://testnet-explorer.gochain.io/tx/0x9d6cb6474e4335d3c929a6039f9c9c9899e5735102701d43536de0d3b911c59f)
For each log, we can inspect `topics[0]` for an eventID to see if it's an event we understand, and also the contract address to see what type of contract it is (necessary since e.g. 20 and 721 Transfer events look the same, but are structured differently due to indexed fields). If we _are_ able to parse the event, then the user should have the option of viewing them that way. However, adding a simple toggle switch like for input doesn't really work since there may be multiple different events, and the events don't contain all the useful metadata (only `topics` and `data`). Instead, we should consider displaying them as a list of separate log entries (still showing the raw JSON for each), and then we can extend/modify the entries for known events with the parsed data.


jmank88:
We could probably just do:

| Index        | Contract        | Topics+Data/Event |
| ------------- | ------------- | ----- |
| 0    |  [`0x12354`]() | ... |
| 0    |  [`0x12354`]() | `Transfer(from: 0x1, to: 0x2, value: 100)` |

And gray out rows for 'removed'. All the other fields are redundant.

jmank88:
The table view can be an additional alternate receipts view behind a switch, just like for input.

bergauz:
Screenshot 2019-10-10 at 23 42 06
Screenshot 2019-10-10 at 23 42 02

front/src/app/scenes/transaction/transaction.component.ts
```diff
@@ -63,6 +70,46 @@ export class TransactionComponent implements OnInit, OnDestroy {
...
+      this._commonService.getAddress(this.tx.to),
+    ]).subscribe(([events, address]: [ContractEventsAbi, Address]) => {
+      this.tx.addressData = address;
+      this.tx.processedLogs = this.tx.parsedLogs.map((log: TxLog) => {
+        const prettyLog: ProcessedLog = new ProcessedLog();
+        prettyLog.index = +log.logIndex;
+        prettyLog.contract_address = log.address;
+        prettyLog.removed = log.removed;
+        let abiItem: AbiItem;
+        const eventSignature = <string>log.topics[0];
```
jmank88: Probably need to account for logs with no topics and not index-out-of-bounds here.

front/src/app/models/transaction.model.ts
```diff
@@ -1,3 +1,17 @@
...
+
+export class ProcessedLog {
+  index: number;
+  contract_address: string;
+  data: string;
+  removed: boolean;
+}
+
+export interface TxLog extends Log {
+  removed: boolean;
```
jmank88: Strange that this field is missing from the official library...

front/src/assets/styles/scss/_variables.scss
```diff
@@ -438,7 +438,7 @@ $btn-box-shadow:              inset 0 1px 0 rgba($white, .15), 0 1px 1px rgba($b
 $btn-focus-width:             $input-btn-focus-width !default;
 $btn-focus-box-shadow:        $input-btn-focus-box-shadow !default;
 $btn-disabled-opacity:        .65 !default;
-$btn-active-box-shadow:       inset 0 3px 5px rgba($black, .125) !default;
+$btn-active-box-shadow:       nonde !default;
```
jmank88: none?

front/src/assets/styles/scss/mixins/_buttons.scss
```diff
@@ -49,14 +49,14 @@
...
-      }
-    }
+    //&:focus {
+    //  // Avoid using mixin so we can pass custom focus shadow properly
+    //  @if $enable-shadows and $btn-active-box-shadow != none {
+    //    box-shadow: $btn-active-box-shadow, 0 0 0 $btn-focus-width rgba(mix(color-yiq($background), $border, 15%), .5);
+    //  } @else {
+    //    box-shadow: 0 0 0 $btn-focus-width rgba(mix(color-yiq($background), $border, 15%), .5);
+    //  }
+    //}
```
jmank88: Remove?

front/src/app/scenes/transaction/transaction.component.html
```diff
@@ -74,20 +74,49 @@
...
           <div class="form-group">
             <label for="receipt_logs">Receipt Logs</label>
-            <textarea
-                class="form-control code"
-                id="receipt_logs"
-                readonly
-                rows="5"
-            >{{tx.logs}}</textarea>
+            <div class="btn-group btn-group-sm float-right">
+              <button class="btn btn-primary" [ngClass]="{'active': !showLogsRaw}" (click)="showLogsRaw = false">Parsed</button>
```
jmank88: `Table`?

front/src/app/scenes/transaction/transaction.component.html
```diff
@@ -74,20 +74,49 @@
...
+                  <tr>
+                    <th>Index</th>
+                    <th>Contract</th>
+                    <th>Log Data</th>
+                  </tr>
+                  </thead>
+                  <tbody>
+                  <tr *ngFor="let log of tx.processedLogs" [ngClass]="{'table-secondary': log.removed}">
+                    <td>{{log.index}}</td>
+                    <td>{{log.contract_address}}</td>
```
jmank88: This can be a link. Also, IMO using a fixed width font looks good for hex addresses.

front/src/app/scenes/transaction/transaction.component.ts
```diff
@@ -63,6 +70,46 @@ export class TransactionComponent implements OnInit, OnDestroy {
...
+        }
+        if (abiItem) {
+          log.topics.shift();
+          const decoded = this._walletService.w3.eth.abi.decodeLog(
+            abiItem.inputs,
+            log.data.replace('0x', ''),
+            <string[]>log.topics
+          );
+          const res: string[] = abiItem.inputs.map(input => `${input.name}: ${decoded[input.name]}`);
+          prettyLog.data = `${abiItem.name}(${res.join(', ')})`;
```
jmank88: It would be neat if values were links, but that might be messy, so it could wait for later.

front/src/app/scenes/transaction/transaction.component.ts
```diff
@@ -63,6 +70,46 @@ export class TransactionComponent implements OnInit, OnDestroy {
...
+    ]).subscribe(([events, address]: [ContractEventsAbi, Address]) => {
+      this.tx.addressData = address;
+      this.tx.processedLogs = this.tx.parsedLogs.map((log: TxLog) => {
+        const prettyLog: ProcessedLog = new ProcessedLog();
+        prettyLog.index = +log.logIndex;
+        prettyLog.contract_address = log.address;
+        prettyLog.removed = log.removed;
+        let abiItem: AbiItem;
+        const eventSignature = <string>log.topics[0];
+        const knownEvent = events[eventSignature];
```
jmank88: For a future enhancement, we can check for and use a verified ABI.

front/src/app/scenes/transaction/transaction.component.ts
jmank88: `address` might be easy to do now, since they are all the same. But some values require extra mapping, e.g. ERC20 Transfer uint256 is a value so a link wouldn't make sense, but ERC721 Transfer uint256 is a token ID and can be a special type of link.

front/src/app/scenes/transaction/transaction.component.html
jmank88: Updated to link.

front/src/app/models/transaction.model.ts
bergauz: yeah, no typo for that

front/src/assets/styles/scss/mixins/_buttons.scss
bergauz: it's styles core, maybe we will need it later

jmank88: Then can you add a note about why it's commented out, so someone who finds this later will know?

bergauz: ok

bergauz:
Screenshot 2019-10-11 at 17 52 39

jmank88:
Looks good! We can probably treat the empty address special by not linking, and could even show `mint`/`burn` in addition to, or in place of it.

PR #415, opened 2019-10-10
labels: bold; drop colon

jmank88:
This PR updates label styles and formatting.

### DIFF
--- a/front/src/assets/styles/scss/_variables.scss
+++ b/front/src/assets/styles/scss/_variables.scss
@@ -289,7 +289,7 @@ $font-size-sm:                $font-size-base * .875 !default;
 $font-weight-lighter:         lighter !default;
 $font-weight-light:           300 !default;
 $font-weight-normal:          400 !default;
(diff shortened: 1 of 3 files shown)


front/src/assets/styles/scss/_variables.scss
```diff
@@ -289,7 +289,7 @@ $font-size-sm:                $font-size-base * .875 !default;
 $font-weight-lighter:         lighter !default;
 $font-weight-light:           300 !default;
 $font-weight-normal:          400 !default;
-$font-weight-bold:            500 !default;
+$font-weight-bold:            600 !default;
```
jmank88: 500 wasn't rendering any differently than 400. `<dt>` depends on this, as well as a few other things.

r-gochain:
@jmank88 could you resolve the conflict

jmank88:
Actually let's close this b/c we're going in a different direction. We can reuse the branch in a new PR.

Issue #274, opened 2019-04-24
Upgradeable Contracts


jmank88:
The web3 upgradeable interfaces ([`package assets`](https://godoc.org/github.com/gochain-io/web3/assets)`UpgradeableProxyABI`, `OwnerUpgradeableProxyABI`) should be automatically detected (related #225, #255).

jmank88:
It would be useful if the result of `target()` was automatically displayed as a link to the target contract page.

jmank88:
Interfaces implemented by target contracts should be automatically detected like normal, but we should consider *blocking* interaction from explorer when we know that a contract is actually the target of an upgradable proxy, since the target state is not valid. Additionally, we should *enable* interaction from the proxy, even though it doesn't directly implement the interfaces.

@benbjohnson Am I getting this right?

benbjohnson:
Yes, if we can block the target contract interactions that would be good but we don't currently have a flag indicating that it is a target.

The ABI is totally opaque to the proxy since it just uses the fallback function to pass along to the target contract. Functions in the ABI are assigned a 4-byte hash so if that hash doesn't match any functions in the proxy contract (e.g. `upgrade()`) then it sends it along as the argument to the target via `delegatecall`.

jmank88:
> Yes, if we can block the target contract interactions that would be good but we don't currently have a flag indicating that it is a target.

Yeah I was having second thoughts about this. Explorer can figure out if a proxy targets it - maybe then it  should have a warning like: "This contract is targeted by upgradable proxy 0x1234"

PR #423, opened 2019-10-15
add target and owner to contract address pages

jmank88:
This PR adds `target` and `owner` fields to the address page for appropriate contracts.

Towards #274


front/src/app/scenes/address/address.component.html
```diff
@@ -8,14 +8,22 @@
...
+                <span class="text-monospace">{{addr.address}}</span>
                 <ng-container *ngIf="addr.signerDetails">
                   (<a [href]="addr.signerDetails.url" target="_blank">{{addr.signerDetails.name}}</a>)
                 </ng-container>
               </dd>
               <dt>Balance:</dt>
               <dd>{{addr.balance_wei | weiToGO | bigNumber}}</dd>
               <ng-container *ngIf="addr.contract">
+                <ng-container *ngIf="addr.target">
+                  <dt>Target:</dt>
```
jmank88: There will be an `Upgradeable` tag nearby, but is there a better way to communicate what this is? Would a little `?` icon tooltip be helpful? (e.g. `Upgradeable contracts execute the code which they target`)

server/tokens/tokens.go
```diff
@@ -154,6 +159,17 @@ func (tb *TokenDetails) queryTokenDetails(conn *goclient.Client, byteCode string
 	lgr.Info("Analyzing contract byte code", zap.Stringer("address", tb.Contract),
 		zap.Strings("ercTypes", tb.ERCTypesSlice()), zap.Strings("functions", tb.FunctionsSlice()),
 		zap.Int("cachedContracts", count))
+	if _, ok := tb.ErcTypes[utils.Upgradeable]; ok {
+		if err := tb.queryTarget(conn, lgr); err != nil {
+			return err
+		}
+		//TODO ensure target loaded in db, and merge its types and funcs in to this set
```
jmank88: To finish #274, we need a hook here to get the types and funcs from the target (which may require a full load, if it's not in the DB), and then merge them in to this one. Then the other checks and the front end will include the full type and func sets. We can add this feature in a future PR.

front/src/app/scenes/address/address.component.html
r-gochain: Let's merge and then update this in a separate PR

PR #426, opened 2019-10-19
parse and format input data

### LINKED ISSUE #281: Parse Tx Input Data
Tx input data is currently viewable as either HEX or UTF-8. When we have validated contract code we should be able to (1) parse and display contract creations as solidity-ish code (with constructor args, if present), and if we have the ABI (or the user enters one) then we should (2) parse and display smart contract calls as solidity-ish code (with human readable arguments, e.g. bytes as hex).

jmank88:
This PR adds the ability to parse and format tx input data, and adds more interface and function definitions.
Screenshot from 2019-10-19 08-33-14
Screenshot from 2019-10-19 08-33-20
Screenshot from 2019-10-19 08-33-29
Screenshot from 2019-10-19 08-34-13
Screenshot from 2019-10-19 08-34-22

Fixes #281


jmank88:
Screenshot from 2019-10-20 09-00-43
CIDs are parsed and formatted now, but the import is sort of hacky.

front/src/app/scenes/transaction/transaction.component.ts
```diff
@@ -2,22 +2,26 @@
...
-import {AbiItem} from 'web3-utils';
+import {AbiItem, AbiInput} from 'web3-utils';
 import {Address} from '../../models/address.model';
+import {Contract} from '../../models/contract.model';
 /*UTILS*/
 import {AutoUnsubscribe} from '../../decorators/auto-unsubscribe';
 import {META_TITLES} from '../../utils/constants';
+import {ErcName} from "../../utils/enums";
+
+import CID from "../../../../node_modules/cids/dist/index.min.js"
```
jmank88: Any suggestions for a better way to do this?

front/src/app/scenes/transaction/transaction.component.ts
```diff
@@ -74,85 +79,205 @@ export class TransactionComponent implements OnInit, OnDestroy {
...
+          item.linkExternal = true;
+          return item;
+        }
+      }
+      if (input.name === 'cid' && input.type === 'bytes' && !input.indexed) {
+        const hex = decoded[input.name];
+        if (hex.length > 2) {
+          try {
+            const cid: CID = new CID(Buffer.from(hex.slice(2), 'hex'));
+            const str = cid.toString();
```
jmank88: We can also use this lib in a few other places to eliminate some service calls.

jmank88:
Improved token values and empty addresses:
Screenshot from 2019-10-22 08-00-26
Screenshot from 2019-10-22 08-00-31

front/src/app/scenes/transaction/transaction.component.ts
bergauz: @jmank88 const CID = require('cids') from docs, i've checked there's no typing for cids

front/src/app/scenes/transaction/transaction.component.html
```diff
@@ -73,33 +73,67 @@
...
+                <b>{{tx.processedInputData?.title}}</b>
+                <ng-container *ngIf="tx.processedInputData?.items?.length > 0">
+                  <ul>
+                    <li *ngFor="let item of tx.processedInputData.items">
+                      <ng-container *ngIf="item.name">
+                        {{item.name}}:
+                      </ng-container>
+                      <ng-container *ngIf="item.link; else regularItem">
+                        <ng-container *ngIf="item.linkExternal; else linkInternal">
+                          <a class="text-monospace" [href]="item.link">
```
bergauz: target="_blank" for a new tab?

front/src/app/scenes/transaction/transaction.component.ts
```diff
@@ -74,85 +79,235 @@ export class TransactionComponent implements OnInit, OnDestroy {
...
+        item.link = decoded[input.name];
+        item.linkExternal = true;
+        return item;
+      }
+    }
+    if (input.name === 'cid' && input.type === 'bytes' && !input.indexed) {
+      const hex = decoded[input.name];
+      if (hex.length > 2) {
+        try {
+          const cid: CID = new CID(Buffer.from(hex.slice(2), 'hex'));
```
bergauz: seems like second param need to be int

bergauz: actually ide was showing me that second param is biteOffset but it encoding so it's good

front/src/app/scenes/transaction/transaction.component.html
jmank88:
```suggestion
                          <a class="text-monospace" [href]="item.link" target="_blank">
```

PR #428, opened 2019-10-25
remove unused fields; docs

jmank88:
This PR removes `Nonce` from the block page (it is only populated for votes, and already interpreted elsewhere), and does some other cleanup.
Screenshot from 2019-10-25 08-21-02

### DIFF
--- a/front/src/app/models/block.model.ts
+++ b/front/src/app/models/block.model.ts
@@ -1,11 +1,12 @@
 import {SignerDetails} from './signer-node';
 
 export class Extra {
-  auth: boolean;
   vanity: string;
-  has_vote: boolean;
-  candidate: string;
-  is_voter_election: boolean;
+
+  has_vote: boolean; // If true, then the remaining vote fields are included.
+  auth: boolean; // Whether voting to authorize (add) or deauthorize (remove).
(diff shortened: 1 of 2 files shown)


front/src/app/models/block.model.ts
```diff
@@ -1,12 +1,13 @@
...
-  has_vote: boolean;
-  candidate: string;
-  is_voter_election: boolean;
-  signerDetails?: SignerDetails;
+
+  has_vote: boolean; // If true, then the remaining vote fields are included.
+  auth: boolean; // Whether voting to authorize (add) or deauthorize (remove).
+  is_voter_election: boolean; // Whether voting on voter (true) or signer (false) auth.
+  candidate: string; // Candidate address.
+  candidateDetails?: SignerDetails;
```
jmank88: I was confusing myself again with these, so I reorganized them. Seems like a `vote?: Vote;` might be a little more intuitive, but this is in the API, so nbd to leave as is.

r-gochain: I think you have to update it on the server side too - https://github.com/gochain/explorer/blob/master/server/models/block.go#L44

jmank88: oh yeah whoops. Better just leave it as it was

PR #430, opened 2019-10-28
metamask support

bergauz:


front/src/app/services/wallet.service.ts
```diff
@@ -17,30 +17,116 @@ import {Transaction} from '../models/transaction.model';
...
+  get ready$(): Observable<boolean> {
+    return this._ready$.pipe(
+      filter<boolean>(v => !!v),
+      take(1),
+    );
+  }
+
   private _web3Callable$: BehaviorSubject<Web3> = new BehaviorSubject(null);
-  private _web3Payable$: BehaviorSubject<Web3> = new BehaviorSubject(null);
+  private _web3Metamask$: BehaviorSubject<Web3> = new BehaviorSubject(null);
```
jmank88: Switching back to Metamask specific naming seems backwards. Why don't we keep these general, for example:
```
walletInstalled$:
walletConfigured$:
_web3Payable$:
```
Then if/when we switch or add support for others, none of this needs to be updated. 
(The user messages can stay metamask specific until we actually add more)

jmank88:
Configuratiron cases to test:
- [ ] MM
- [ ] PK w/o MM
- [ ] PK w/ MM
- [ ] PK w/ MM misconfigured (different net id)

bergauz:
@jmank88 if MM available there's no option to enter PK

front/src/app/services/wallet.service.ts
bergauz: @jmank88 can you check new code, wallet and metamask is different, pk is also wallet

treeder:
What are we doing with this?

bergauz:
@treeder pushed phase 1 from #440 , have some questions, should user enter PK each time when sending transaction?

treeder:
No, if they stay on the page, leave the pk in the field.

PR #438, opened 2019-10-29
download bytecode when contract is not yet discovered

r-gochain:


server/backend/backend.go
```diff
@@ -208,7 +209,23 @@ func (self *Backend) GetContract(contractAddress string) (*models.Contract, erro
...
+		contractDataArray, err := self.CodeAt(context.Background(), normalizedAddress)
+		if err != nil {
+			return nil, err
+		}
+		contractData := string(contractDataArray[:])
+		if contractData == "" {
+			return nil, fmt.Errorf("invalid contract: %s", contractAddress)
+		}
+		byteCode := hex.EncodeToString(contractDataArray)
+		self.ImportContract(normalizedAddress, byteCode)
```
jmank88: Should an error here at least be logged?

server/backend/backend.go
```diff
@@ -208,7 +209,23 @@ func (self *Backend) GetContract(contractAddress string) (*models.Contract, erro
 	if !common.IsHexAddress(contractAddress) {
 		return nil, fmt.Errorf("invalid hex address: %s", contractAddress)
 	}
-	return self.mongo.getContract(common.HexToAddress(contractAddress).Hex())
+	normalizedAddress := common.HexToAddress(contractAddress).Hex()
+	contract, err := self.mongo.getContract(normalizedAddress)		
+	if err != nil || contract==nil {
```
jmank88: Does it make sense to do this for errors? Or only for 'not found'? We return `nil, nil` for 'not found', and I would think that any error would just mean more errors below, so we could just return the first error (or retry?).

jmank88: Could do this, and then drop the indent for this block:
```go
if err != nil || contract != nil {
  return contract, err
}
...
```

Issue #421, opened 2019-10-14
Allow admin to force reload


r-gochain:
Allow admin to force reload

- Block
- Transaction
- Tokens of the contract

r-gochain:
@jmank88 @bergauz any other thoughts?

PR #442, opened 2019-11-02
Admin-cli for explorer

r-gochain:
This is a draft of admin-cli related to this issue - #421, the cli tool should be run on the same instance as explorer inside the container (using either a separate container or the same one) to have direct access to mongodb.


r-gochain:
@jmank88 @treeder PTAL

server/utils/uniq.go
```diff
@@ -0,0 +1,11 @@
+package utils
+
+//AppendIfMissing appends to an array only uniq elements
+func AppendIfMissing(slice []string, i string) []string {
```
jmank88: It looks like all the usages of this func are calls from loops, which is `O(n^2)` overall. It would be more efficient to build `map[string]struct{}`s along the way, then iterate once at the end to make the slice if necessary.

server/backend/db_backend.go
```diff
@@ -35,7 +35,8 @@ type MongoBackend struct {
 // New create new rpc client with given url
 func NewMongoClient(client *goclient.Client, host, dbName string, lgr *zap.Logger) (*MongoBackend, error) {
 	session, err := mgo.DialWithInfo(&mgo.DialInfo{
-		Addrs: []string{host},
+		Addrs:   []string{host},
+		Timeout: 10 * time.Second,
```
jmank88: Is this when the client quits on a long query?

admin/main.go
```diff
@@ -0,0 +1,305 @@
...
+		contract = true
+		byteCode := hex.EncodeToString(contractDataArray)
+		if err := backendInstance.ImportContract(normalizedAddress, byteCode); err != nil {
+			fatalExit(fmt.Errorf("failed to import contract: %v", err))
+		}
+		contractFromDB, err := backendInstance.GetAddressByHash(ctx, normalizedAddress)
+		if err != nil {
+			fatalExit(fmt.Errorf("failed to get contract from DB: %v", err))
+		}
+		fromBlock, err := backendInstance.GetContractBlock(normalizedAddress)
```
jmank88: Check this error?

admin/main.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	balance, err := backendInstance.Balance(ctx, addr)
+	if err != nil {
+		fatalExit(fmt.Errorf("failed to get balance"))
+	}
+	currentBlock, err := backendInstance.GetLatestBlockNumber(ctx)
+	if err != nil {
+		logger.Error("Update Addresses: Failed to get latest block number", zap.Error(err))
+	}
+	contractDataArray, err := backendInstance.CodeAt(ctx, normalizedAddress)
```
jmank88: Check error

server/backend/db_backend.go
r-gochain: both (unfortunately without timeout it could stuck during start, but in the same time it also throwing timeout during long running queries - which is I won't say bad)

r-gochain: set to 2 minutes

PR #443, opened 2019-11-10
Optimizing transactions collection

r-gochain:
- [x] Grabber writing to both collections
- [x] DB migration
- [x] Explorer using both collections (switching to the second one after full migration)


server/utils/migrate.go
```diff
@@ -0,0 +1,195 @@
+//from https://raw.githubusercontent.com/appleboy/mgo-migrate/master/migrate.go had to replace imports
+package utils
```
jmank88: Maybe this should be in its own `migrate` package? And could be under an `internal/` too so we aren't exporting it outside the repo.

jmank88: Would be good to reference the exact version copied too, instead of `master`.

server/backend/db_migrations.go
```diff
@@ -0,0 +1,49 @@
...
+				if err != nil {
+					return err
+				}
+				_, err = self.mongo.C("ExpandedTransactions").Upsert(bson.M{"address": tx.To, "tx_hash": tx.TxHash},
+					bson.M{"address": tx.To, "tx_hash": tx.TxHash, "created_at": tx.CreatedAt})
+				if err != nil {
+					return err
+				}
+			}
+			return nil
```
jmank88:
```suggestion
			return txs.Err()
```

server/backend/db_migrations.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (self *MongoBackend) Migrate(ctx context.Context, lgr *zap.Logger) error {
+	m := migrate.New(ctx, self.mongo, lgr, migrate.DefaultOptions, []*migrate.Migration{{
+		ID:      1,
+		Comment: "Creating ExpandedTransactions collection",
+		Migrate: func(ctx context.Context, d *mgo.Database, lgr *zap.Logger) error {
+			counter := 0
+			var tx *models.Transaction
+			find := d.C("Transactions").Find(bson.M{})
+			txs := find.Iter()
```
jmank88:
```suggestion
			txs := find.Iter()
			defer txs.Close()
```

server/backend/db_migrations.go
```diff
@@ -0,0 +1,49 @@
...
+	migrate "github.com/gochain-io/explorer/server/utils"
+	"go.uber.org/zap"
+	mgo "gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+)
+
+func (self *MongoBackend) Migrate(ctx context.Context, lgr *zap.Logger) error {
+	m := migrate.New(ctx, self.mongo, lgr, migrate.DefaultOptions, []*migrate.Migration{{
+		ID:      1,
+		Comment: "Creating ExpandedTransactions collection",
```
jmank88: What about a more intuitive name, like `TransactionAddresses` or `TranscationsByAddress`?

server/backend/db_backend.go
```diff
@@ -238,6 +247,15 @@ func (self *MongoBackend) importTx(ctx context.Context, tx *types.Transaction, b
 		return fmt.Errorf("failed to upsert tx: %v", err)
 	}
 
+	err = self.insertTransactionsByAddress(ctx, transaction.From, transaction.TxHash, transaction.CreatedAt)
+	if err != nil {
+		return err
+	}
+	err = self.insertTransactionsByAddress(ctx, transaction.To, transaction.TxHash, transaction.CreatedAt)
```
jmank88: What if `to == from`? What if `to == 0x0`?

server/backend/db_backend.go
```diff
@@ -554,27 +580,52 @@ func (self *MongoBackend) getTransactionList(
...
+				"$gte": filter.FromTime,
+				"$lte": filter.ToTime,
+			},
+		}
+		err := self.mongo.C("TransactionsByAddress").
+			Find(findQuery).
+			Sort("-created_at").
+			Skip(filter.Skip).
+			Limit(filter.Limit).
+			All(&transactionsList)
```
jmank88: Can't we join this query against the `Transactions` collection to do it all at once? Or at least fetch them all in a batch to do it all in 2 calls?

server/backend/db_backend.go
```diff
@@ -824,3 +875,11 @@ func (self *MongoBackend) cleanUp() {
 		self.Lgr.Info("Cleaned collection", zap.String("collection", collectionName))
 	}
 }
+func (self *MongoBackend) useTransactionsByAddress() bool {
+	v, err := self.getDatabaseVersion()
+	if err != nil {
+		self.Lgr.Error("Cannot get database version", zap.Error(err))
+		return false
+	}
+	return v > 0
```
jmank88: It might be good to tie this to the migration ID directly, e.g. `return v >= migrationIDTransactionsByAddress`, or with more refactoring `return v >= migrationTransactionsByAddress.ID`.

server/backend/db_backend.go
r-gochain: That could be tricky in mongo, but I'll run some tests with some mongo js (you could iterate through results inside mongo query and in theory it could be faster than golang->mongo->golang->mongo if we won't hit any suboptimal problems)

server/backend/db_backend.go
r-gochain: to==from - only one transaction would be referred (we're using this collection only for extracting IDs nothing else) - there is upsert and uniq index on tx_hash + address

to==0x0 - fixed

jmank88: I know it will just overwrite and not produce a duplicate, but why make the extra call at all, when we could just do a simple check for `to == from` and skip when it's unnecessary?

server/backend/db_backend.go
jmank88: It looks like mongo supports types joins natively. Is this feature available for us? https://docs.mongodb.com/manual/reference/operator/aggregation/lookup/

r-gochain: Yeah looking into it (should be there), hope it would be faster

server/backend/db_backend.go
r-gochain: makes sense

grabber/main.go
```diff
@@ -144,6 +145,14 @@ func main() {
 	}
 	logger.Info("Stopping")
 }
+func migrator(ctx context.Context, importer *backend.Backend, lgr *zap.Logger) {
+	err := importer.MigrateDB(ctx, lgr)
+	if err != nil {
+		importer.Lgr.Fatal("Migration failed", zap.Error(err))
```
jmank88: Is this truly fatal? Wouldn't we prefer to have the old format continue to be imported rather than go in to a restart-crash loop where no new data is imported?

jmank88: And should both log lines be using the same logger?

server/backend/db_backend.go
```diff
@@ -25,11 +25,12 @@ import (
...
-	mongo        *mgo.Database
-	mongoSession *mgo.Session
-	goClient     *goclient.Client
-	Lgr          *zap.Logger
+	host            string
+	mongo           *mgo.Database
+	mongoSession    *mgo.Session
+	goClient        *goclient.Client
+	Lgr             *zap.Logger
+	databaseVersion int
```
jmank88: Accessing this field is racey. It needs to be protected by a `sync.RWMutex`, or accessed via `sync/atomic` funcs (which may also require special placement within the struct fields?).

r-gochain: yep, added mutex and removed updating this field in migrator (unnecessary since it's in grabber)

r-gochain:
Before rolling out to testnet need to increase storage size.

server/backend/db_backend.go
```diff
@@ -46,6 +49,7 @@ func NewMongoClient(client *goclient.Client, host, dbName string, lgr *zap.Logge
 	importer.mongoSession = session
 	importer.mongo = session.DB(dbName)
 	importer.goClient = client
+	importer.databaseVersion = 0
```
jmank88:
```suggestion
```

server/backend/db_migrations.go
```diff
@@ -0,0 +1,71 @@
...
+		ID int `bson:"ID"`
+	}
+	err := self.mongo.C(migrationCollection).Find(nil).Sort("-ID").One(&result)
+	if err != nil {
+		return 0, err
+	}
+	self.mutex.Lock()
+	self.databaseVersion = result.ID
+	self.mutex.Unlock()
+	return self.databaseVersion, nil
```
jmank88:
```suggestion
	return result.ID, nil
```

server/backend/db_migrations.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"go.uber.org/zap"
+	mgo "gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+)
+
+var migrationCollection = "Migrations"
+
+func (self *MongoBackend) getDatabaseVersion() (int, error) {
+	if self.databaseVersion != 0 {
```
jmank88: All accesses to `databaseVersion` need to lock the mutex.

server/backend/db_backend.go
```diff
@@ -554,27 +584,60 @@ func (self *MongoBackend) getTransactionList(
...
+			{"$match": findQuery},
+			{"$lookup": bson.M{
+				"from":         "Transactions",
+				"localField":   "tx_hash",
+				"foreignField": "tx_hash",
+				"as":           "tx",
+			}},
+			{"$unwind": bson.M{
+				"path": "$tx",
+			}},
```
jmank88: Could we add something like `{ $replaceRoot: { newRoot: "$tx" } }` here? Then we can drop the extra types and just use `models.Transaction` directly.

server/backend/db_backend.go
```diff
@@ -25,11 +26,13 @@ import (
...
-	mongo        *mgo.Database
-	mongoSession *mgo.Session
-	goClient     *goclient.Client
-	Lgr          *zap.Logger
+	host            string
+	mongo           *mgo.Database
+	mongoSession    *mgo.Session
+	goClient        *goclient.Client
+	Lgr             *zap.Logger
+	mutex           sync.RWMutex
```
jmank88: A name like `databaseVersionMu` would be more idiomatic and clear about what it applies to.

server/backend/db_migrations.go
```diff
@@ -0,0 +1,74 @@
...
+	mgo "gopkg.in/mgo.v2"
+	"gopkg.in/mgo.v2/bson"
+)
+
+var migrationCollection = "Migrations"
+
+func (self *MongoBackend) getDatabaseVersion() (int, error) {
+	self.databaseVersionMutex.Lock()
+	version := self.databaseVersion
+	self.databaseVersionMutex.Unlock()
```
jmank88: Only need `RLock()/RUnlock()` here

server/models/transactions_by_address.go
```diff
@@ -0,0 +1,9 @@
+package models
+
+import "time"
+
+type TransactionsByAddress struct {
```
jmank88: Is this unused now?

r-gochain: still used on inserts (grabber writes to both collections - Transactions and TransactionsByAddress)

jmank88: Is it? I can't find a usage. The upsert is using `bson.M`.

r-gochain: Yep you're right, I'm using raw bson instead of model.
Removed

PR #448, opened 2020-01-07
Fix extra field and provider names

r-gochain:


server/backend/backend.go
```diff
@@ -470,15 +470,15 @@ func fillExtra(block *models.Block) *models.Block {
...
+	block.Extra.Vanity = utils.CleanUpText(string(clique.ExtraVanity(extra)))
 	block.Extra.HasVote = clique.ExtraHasVote(extra)
 	block.Extra.Candidate = clique.ExtraCandidate(extra).String()
 	block.Extra.IsVoterElection = clique.ExtraIsVoterElection(extra)
 	return block
 }
 func fillExtraLight(block *models.LightBlock) *models.LightBlock {
 	extra := []byte(block.ExtraData)
-	block.Extra.Vanity = string(clique.ExtraVanity(extra))
+	block.Extra.Vanity = utils.CleanUpText(string(clique.ExtraVanity(extra)))
```
jmank88: I'm not sure we should be modifying the whole value, or looking for runes other than `\u0000` - those seems like separate problems, if they are even real problems at all. The trailing 0s on the other hand are explicitly expected, and much simpler to handle - we only need something like: `strings.TrimRight(extra, "\u0000")`, which could be inlined at these two call sites.

PR #450, opened 2020-01-20
support solc evm-version when verifying contracts

jmank88:
Adds a dropdown to the UI to optionally select and EVM version:
Screenshot from 2020-01-27 09-31-47
Example request:
Screenshot from 2020-01-27 09-31-56
`Default` is blank string, which uses the old behavior of no flag, so all the existing entries are forward compatible.

### DIFF
--- a/front/src/app/scenes/contract/contract.component.html
+++ b/front/src/app/scenes/contract/contract.component.html
@@ -47,6 +47,21 @@
             <option *ngFor="let compiler of compilers">{{compiler}}</option>
           </select>
         </div>
+        <div class="form-group">
+          <label for="evm">EVM</label>
+          <select
+              class="form-control"
+              id="evm"
+              formControlName="evm_version">
+            <option value="" selected>Default</option>
+            <option value="homestead">Homestead</option>
+            <option value="tangerineWhistle">Tangerine Whistle</option>
+            <option value="spuriousDragon">Spurious Dragon</option>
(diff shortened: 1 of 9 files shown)


front/src/app/scenes/contract/contract.component.html
```diff
@@ -47,6 +47,21 @@
             <option *ngFor="let compiler of compilers">{{compiler}}</option>
           </select>
         </div>
+        <div class="form-group">
+          <label for="evm">EVM</label>
+          <select
+              class="form-control"
+              id="evm"
+              formControlName="evm_version">
+            <option value="" disabled selected>Select EVM</option>
```
jmank88: This is optional, so should this option just be `Default`?

front/src/app/scenes/contract/contract.component.html
```diff
@@ -47,6 +47,21 @@
             <option *ngFor="let compiler of compilers">{{compiler}}</option>
           </select>
         </div>
+        <div class="form-group">
+          <label for="evm">EVM</label>
+          <select
+              class="form-control"
+              id="evm"
+              formControlName="evm_version">
```
r-gochain: what it affects? or you haven't added a code yet?

jmank88: This should be working end-to-end. `evm_version` is passed along through the backend to the `solc` docker command.

PR #462, opened 2020-05-16
Attempt to fix #452 - Decimals not taken into account on token balances

### LINKED ISSUE #452: Decimals not taken into account on token balances
Perhaps elsewhere too.

This contract has 6 decimals and it's showing at zero value when it should be 1. 

image

treeder:


r-gochain:
image

r-gochain:
@treeder please fix the tests

./main.go:130:38: not enough arguments in call to backend.NewBackend
 
	have (context.Context, string, string, string, []string, nil, *zap.Logger)
 
	want (context.Context, string, string, string, []string, map[common.Address]models.Signer, *zap.Logger, *ristretto.Cache)
 
note: module requires Go 1.14
 
make: *** [Makefile:13: grabber] Error 2

treeder:
Fixed. Sorry about that. BTW, the cache I added there could probably help speed up the grabber and things.

server/backend/backend.go
```diff
@@ -159,6 +179,8 @@ func (self *Backend) GetAddressByHash(ctx context.Context, hash string) (*models
 	}
 	address.BalanceWei = balance.String() //to make sure that we are showing most recent balance even if db is outdated
 	address.BalanceString = new(big.Rat).SetFrac(balance, wei).FloatString(18)
+	// todo: only store a subset of this data in the cache, just the metadata like token name, decimals, etc. Things that won't change.
```
jmank88, line 182: Is this a blocker for this PR? Or is there some kind of default timeout built in to the cache so that we'll at least be eventually consistent?

treeder: Not a blocker since it's only used in that one function for now and it doesn't use the values that can change. If we start using these cache entries for the values that can change, then we'll want to update this.

treeder:
Ehh, sorry, guess this might not all be fixed. When I run `make test-backend`, I don't get these build errors though. Is there a different test command I should be running?

PR #480, opened 2020-12-09
contract verification: improve v29 regexp; add v32 and v33

jmank88:
Contract verification is meant to ignore metadata differences, but our regexp for matching the metadata suffixes was limited to a subset of version 29. This change improves the v29 pattern (more accurate and faster), and adds patterns for v32 and v33.


server/backend/solc.go
```diff
@@ -151,15 +151,28 @@ func ParseCombinedJSON(combinedJSON []byte, source string, languageVersion strin
 	return contracts, nil
 }
 
-var solcMetadataRegex = regexp.MustCompile(`056fea165627a7a72305820.*0029$`)
+var (
+	// a2 65 'bzzr' <1 byte version> 58 20 <32 bytes swarm hash> 00 29
+	solcMetadata29Suffix = regexp.MustCompile(`a165627a7a72.{2}5820.{64}0029$`)
```
jmank88, line 156: The `056fe` before this was overly strict, and the unrestricted length for the hash was looser than necessary (plus fixed length matching should be faster).

treeder (approved):

PR #482, opened 2021-01-07
add gas fees field to blocks; remove eth_totalSupply dependency

jmank88:
This change adds a gas fees field to blocks (as well as some other minor changes) in preparation for calculating total supply after GIP 30.
A follow up change is required to implement the `TotalFeesBurned` calculation, and to pick up the `ChainConfig` after the gochain repo is updated with the fork information. However, we could merge this first since `Gas Fees` is interesting block meta data regardless of whether the fork is applied.

### DIFF
--- a/front/src/app/scenes/block/block.component.html
+++ b/front/src/app/scenes/block/block.component.html
@@ -23,6 +23,10 @@
           <dd>{{block.gas_limit | bigNumber}}</dd>
           <dt>Gas Used</dt>
           <dd>{{block.gas_used | bigNumber}}</dd>
+          <ng-container *ngIf="block.gas_fees">
+            <dt>Gas Fees</dt>
+            <dd>{{block.gas_fees | weiToGO | bigNumber}}</dd>
+          </ng-container>
           <dt>Transactions Count</dt>
           <dd>{{block.tx_count}}</dd>
           <dt>Difficulty</dt>
--- a/grabber/main.go
+++ b/grabber/main.go
@@ -39,7 +39,7 @@ func main() {
 	}()
 
 	var rpcUrl string
-	var checkTxCount bool
+	var checkExternal bool
 	var mongoUrl string
 	var dbName string
 	var startFrom int64
@@ -64,8 +64,8 @@ func main() {
 		},
 		cli.BoolFlag{
 			Name:        "tx-count, tx",
-			Usage:       "check a transactions count for every block(a heavy operation)",
-			Destination: &checkTxCount,
+			Usage:       "verify transaction count against RPC for every block(a heavy operation)",
+			Destination: &checkExternal,
 		},
 		cli.StringFlag{
 			Name:        "mongo-dbname, db",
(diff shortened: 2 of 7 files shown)


grabber/main.go
```diff
@@ -328,15 +328,16 @@ func checkAncestors(ctx context.Context, importer *backend.Backend, blockNumber
 }
 
 // ensureTxsConsistent checks if the block tx count matches the number of txs in the db, and reimports the block if not.
-// If a block is reimported, then the parent hash is returned.
-func ensureTxsConsistent(ctx context.Context, importer *backend.Backend, blockNumber int64, checkTxCount bool) (common.Hash, error) {
-	if ok, err := importer.TransactionsConsistent(blockNumber); err != nil {
+// If a block is reimported and the parent changed, then the parent hash is returned.
```
jmank88, line 331: This should be an efficiency improvement, since the sole caller was assuming this already.

server/backend/db_backend.go
```diff
@@ -167,11 +168,36 @@ func (self *MongoBackend) importBlock(ctx context.Context, block *types.Block) (
 	if err != nil {
 		return nil, fmt.Errorf("failed to remove old txs: %v", err)
 	}
+	gasFee := big.NewInt(0)
 	for _, tx := range block.Transactions() {
-		if err := self.importTx(ctx, tx, block); err != nil {
+		if fee, err := self.importTx(ctx, tx, block); err != nil {
 			return nil, fmt.Errorf("failed to import tx: %v", err)
+		} else {
+			gasFee = gasFee.Add(gasFee, fee)
```
jmank88, line 176: Summing up the fees is cheap since we were already looping over the txs.

server/backend/db_backend.go
```diff
@@ -167,11 +168,36 @@ func (self *MongoBackend) importBlock(ctx context.Context, block *types.Block) (
...
+			return nil, err
+		}
+		if parent != nil && parent.TotalFeesBurned != "" {
+			totalFeesBurned, ok := new(big.Int).SetString(parent.TotalFeesBurned, 10)
+			if !ok {
+				lgr.Error("Failed to parse block.TotalFeesBurned as big.Int", zap.String("block", b.ParentHash),
+					zap.String("value", parent.TotalFeesBurned))
+			} else {
+				b.TotalFeesBurned = totalFeesBurned.Add(totalFeesBurned, gasFee).String()
+				update["total_fees_burned"] = b.TotalFeesBurned
```
jmank88, line 193: Tracking the cumulative total is easy in the normal case of chronological import where the parent is on hand already.
During a reverse import, when the parent blocks are not available, we just have to trigger some cleanup routine once the chain is 'connected' again.

front/src/app/scenes/block/block.component.html
```diff
@@ -23,6 +23,10 @@
           <dd>{{block.gas_limit | bigNumber}}</dd>
           <dt>Gas Used</dt>
           <dd>{{block.gas_used | bigNumber}}</dd>
+          <ng-container *ngIf="block.gas_fees">
+            <dt>Gas Fees</dt>
+            <dd>{{block.gas_fees | weiToGO | bigNumber}}</dd>
+          </ng-container>
```
jmank88, line 29: If we pull chain config info to the front end, then we could annotate `Gas Fees` as being transferred to the signer address or burned depending on whether the fork is applied.
Screenshot from 2021-01-07 09-09-46
Screenshot from 2021-01-07 09-10-09

r-gochain (approved):

PR #484, opened 2021-03-08
track burned gas fees for darvaza

jmank88:


jmank88:
I found a pre-existing bug with tx gas fee calculations that was throwing off the blocks indirectly. Txs would initially have their fee calculated from the gas limit, then later overwritten based on the gas used. Should be fixed now.

r-gochain (approved):

jmank88:
I was able to build and run this locally, but not test everything since it would have to sync a ton of data before triggering all the new paths. @r-gochain Does this seems safe enough to merge and release to the testnet?

r-gochain:
Yep let's try on the testnet

PR #490, opened 2021-03-16
remove redundant tx write

jmank88:


server/backend/db_backend.go
```diff
@@ -279,10 +279,6 @@ func (mb *MongoBackend) importTx(ctx context.Context, tx *types.Transaction, blo
 		}
 	}
```
jmank88, line 281: The Upsert below was redundant now that ensureReceipt is always called prior, which does it's own write.

treeder (approved):

PR #519, opened 2021-07-14
don't log fatal for http.ErrServerClosed; warn about close errors

jmank88:
Normal shutdown of the http server returns `http.ErrServerClosed`, which we just see as an error and log fatal:
Screenshot from 2021-07-14 09-08-40
Now we check for `http.ErrServerClosed` and return nil instead.


jmank88:
Another part of normal shutdown is context cancellation which is logged as an 'error' in some cases:
Screenshot from 2021-07-14 09-13-21

r-gochain (approved):

PR #530, opened 2021-09-21
status badge on tx list(gochain#524)

aug2uag:


treeder:
BTW, not sure why this took so long to deploy, but looks awesome, thanks!

image

Issue #541, opened 2021-11-11
All token transfers show failed chip


treeder:
Even if they are successful:

![image](https://user-images.githubusercontent.com/75826/141307004-d1bc26c2-d5d9-4109-b68a-a64700e5b31c.png)

treeder:
@aug2uag fyi

aug2uag:
looking

aug2uag:
@treeder are these any specific kind of transaction? I was looking to see if it's something I did, but it doesn't look like I touched the JS based on git log diffs

treeder:
I'd imagine it's just something you overlooked or something when adding the feature. Just go to any wallet and you'll see it's all red, for example: https://explorer.gochain.io/addr/0xD93e990048581e218182bC3d432d8750B5CDDF4a?addr_tab=token_transactions

aug2uag:
I'm currently unable to find addresses with token/transactions in local dev env. The issue is localized to `token-txs.component.html:67`, needs introspection with accompanying `token-txs.component.ts` file. I'm guessing `status` either doesn't exist (i.e., would need mapping in TS file) or the attribute may be capitalized or similar

Grabber is running, I may need to drop db and seed again

aug2uag:
I can query an address with many transactions, for me found one at `0x09344477fDc71748216a7b8BbE7F2013B893DeF8` but it doesn't have token transfers and there doesn't seem a way to identify token transfers from db:
```
> db.Transactions.findOne()
{
	"_id" : ObjectId("61498153db80562dcce3315c"),
	"tx_hash" : "0xcc9f5740af7b01ad4c53854a5596f664be95b75240c70bb120e0ce8e1fd6de44",
	"to" : "0x61bA2481AA6CF6fA85d88A09A2c65B622553B902",
	"from" : "0x09344477fDc71748216a7b8BbE7F2013B893DeF8",
	"status" : true,
	"contract_address" : "0x0000000000000000000000000000000000000000",
	"value" : "75030000000000000",
	"gas_price" : "2000000000000",
	"gas_fee" : "42000000000000000",
	"gas_limit" : NumberLong(21500),
	"block_number" : NumberLong(21143602),
	"nonce" : NumberLong(25982),
	"hash" : "0x36f41928d1e1a60c2460cf92ea845c9c350ff19077722c4e04ad9c8ace7991a4",
	"created_at" : ISODate("2021-09-21T06:52:59Z"),
	"input_data" : "",
	"logs" : "[]",
[...]

PR #542, opened 2021-11-11
All token transfers show failed chip

aug2uag:
#541


treeder:
@r-gochain are all internal transactions considered successful if they show up here?

r-gochain:
I think so, I just cannot imagine the way how it could fail (it's either
rolled back or successful)

PR #546, opened 2021-12-28
 Support EIP-3091 #525

### LINKED ISSUE #525: Support EIP-3091
It may already be like this, if so close this. 

https://eips.ethereum.org/EIPS/eip-3091

r-gochain:
closes #525

### DIFF
--- a/server/main.go
+++ b/server/main.go
@@ -248,6 +248,12 @@ func main() {
 				})
 
 				r.Get("/contracts", getContractsList)
+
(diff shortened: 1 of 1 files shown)


treeder:
I think the EIP is more for front end, but I suppose it doesn't hurt to match on the backend. Do the front end URLs already support this?

r-gochain:
Works on the frontend as well

PR #551, opened 2022-02-01
fix transaction counter for "to" only addresses

r-gochain:


server/backend/backend.go
```diff
@@ -272,7 +273,14 @@ func (b *Backend) GetAddressByHash(ctx context.Context, hash string) (*models.Ad
 	if err != nil {
 		return nil, err
 	}
-	address.NumberOfTransactions = int(nonce)
+	transactionCounter := int(nonce)
```
treeder, line 276: Is nonce ever the right value?  Incoming transactions won't be counted if we use the nonce right?

r-gochain: Updated, now using for everything

treeder:
Would it make more sense to increment a counter rather than doing a count query every request?

r-gochain:
@treeder If I'm not mistaken that was one of the first version (counter), the problem is that counter was constantly getting off (and it's hard to backfill the data) so we decided to use intermediate collection (TransactionsByAddress)

PR #557, opened 2022-02-28
Bugfix for solc output parsing

r-gochain:


treeder (approved):
Maybe we should say only 0.8+ versions supported on the interface?

r-gochain:
@treeder as an option, but I suppose we should support both (not sure they are backward compatible)

